PR #3, opened 2023-02-20
C++ sanity test

AlexKoff88:


model_api/cpp/install_dependencies.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+#Install OpenCV
+apt-get install libopencv-dev
```
Wovchena, line 4: You may want to put this after `apt update` to ensure the latest registry is used

AlexKoff88: Good catch. Will do this.

Wovchena (approved):

PR #8, opened 2023-02-23
Added CLI params to tests

AlexKoff88:


Wovchena (approved):
Next time I suggest using gflags to parse cmd to stay aligned with demos and samples. But since this is for test, I guess we can go with this for now

PR #9, opened 2023-02-23
Add Model suffix

Wovchena:


Wovchena:
@AlexKoff88 I started to have a doubt if we need this change. Maybe we should wait for Geti to move to this repo first. And after that introduce incompatible changes.

PR #21, opened 2023-03-02
Introduce InferenceAdapter in C++ part

AlexKoff88:
- Add InferenceAdapter interface
- Add OpenVINOInferenceAdapter for local inference
- model->load() method is introduced and called in the .create_model() for smooth UX (see detection_model for details)


Radwan-Ibrahim (approved):
LGTM

model_api/cpp/models/src/model_base.cpp
```diff
@@ -27,28 +28,30 @@
...
+}
 
-std::shared_ptr<ov::Model> ModelBase::prepareModel(ov::Core& core) {
+void ModelBase::prepareModel(ov::Core& core) {
     // --------------------------- Read IR Generated by ModelOptimizer (.xml and .bin files) ------------
     /** Read model **/
-    slog::info << "Reading model " << modelFileName << slog::endl;
-    std::shared_ptr<ov::Model> model = core.read_model(modelFileName);
+    slog::info << "Reading model " << modelFile << slog::endl;
+    std::shared_ptr<ov::Model> model = core.read_model(modelFile);
```
ljcornel, line 46: Would it be possible to add a `InferenceAdapter:getOpenVINOModel` (or something like that) method to the InferenceAdapter interface to avoid calling `core.read_model` here?

ljcornel (approved):
Looks good!

model_api/cpp/adapters/src/openvino_adapter.cpp
```diff
@@ -0,0 +1,73 @@
...
+    // Fill input blobs
+    for (const auto& item : input) {
+        inferRequest.set_tensor(item.first, item.second);
+    }
+
+    // Do inference
+    inferRequest.infer();
+
+    // Processing output blobs
+    //auto output = std::make_unique<InferenceOutput>();
```
Wovchena, line 50:
```suggestion
```

Wovchena (approved):
There are conflicts now

model_api/cpp/models/src/model_base.cpp
AlexKoff88: It means that ov::Model object will live for the whole ModelBase object lifetime and lead to the extra RAM consumption during the inference. We will get rid of this when revise preprocessing embedding step.

PR #30, opened 2023-03-08
Align C++ and Python classification accuracy

Wovchena:
Replace ASSERT_TRUE with ASSERT_EQ because it reports the actual values if there's mismatch.

Detection accuracy still diverges.

### DIFF
--- a/model_api/python/openvino/model_api/models/classification.py
+++ b/model_api/python/openvino/model_api/models/classification.py
@@ -71,7 +71,6 @@ class ClassificationModel(ImageModel):
     @classmethod
     def parameters(cls):
         parameters = super().parameters()
-        parameters["resize_type"].update_default_value("crop")
         parameters.update(
             {
                 "topk": NumericalValue(value_type=int, default_value=1, min=1),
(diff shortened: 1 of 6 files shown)


model_api/python/openvino/model_api/models/utils.py
```diff
@@ -122,15 +122,11 @@ def load_labels(label_file):
...
-        resized_frame = cv2.resize(image, size, interpolation=interpolation)
-    else:
+    if keep_aspect_ratio:
         h, w = image.shape[:2]
         scale = min(size[1] / h, size[0] / w)
-        resized_frame = cv2.resize(
-            image, None, fx=scale, fy=scale, interpolation=interpolation
-        )
-    return resized_frame
+        return cv2.resize(image, None, fx=scale, fy=scale, interpolation=interpolation)
```
AlexKoff88, line 128: We probably need to align preprocessing embedding as well.

model_api/python/openvino/model_api/models/classification.py
```diff
@@ -85,13 +84,13 @@ def parameters(cls):
 
     def postprocess(self, outputs, meta):
         outputs = outputs[self.out_layer_name].squeeze()
+        if not np.isclose(np.sum(outputs), 1.0, atol=0.01):
```
AlexKoff88, line 87: Do you think that this is the right check that reflects what happens in the C++ part?

model_api/python/openvino/model_api/models/utils.py
Wovchena: Removal of `parameters["resize_type"].update_default_value("crop")` affects preprocessing only. The part you left the comment under is just rephrasing, it's effectively the same

model_api/python/openvino/model_api/models/classification.py
Wovchena: C++ embeds softmax into `ov::Model`. We used to have it in Python as well. But we had to remove it to support `OVMSAdapter` in python which doesn't allow to manipulate `ov::Model` on the server side.

model_api/python/openvino/model_api/models/utils.py
AlexKoff88: Ok. I see

PR #37, opened 2023-03-16
Serialize model

Wovchena:


model_api/python/openvino/model_api/models/model.py
```diff
@@ -469,3 +469,11 @@ def log_layers_info(self):
                     name, metadata.shape, metadata.precision, metadata.layout
                 )
             )
+
+    def serialize(self, xml_path, bin_path="", version="UNSPECIFIED"):
```
AlexKoff88, line 473: Can we just rename this method to `.save()`?

This implementation is slightly different from what we agreed to have in C++ API when this should be a `.getModel()` method that just returns a model (with all the properties in model info) and the user can serialize this on his end as he wants. 
Maybe we can go the same way here,  like providing `.get_model()` method which returns the model with filled rt_info and in addition to it `.save()` method which will use  `.get_model()` underneath.
What do you think?

PR #44, opened 2023-03-23
[DRAFT]: .getModel() in C++ and config tests

AlexKoff88:


tests/cpp/precommit/test_model_config.cpp
```diff
@@ -0,0 +1,117 @@
...
+// TODO: Add tests for create_model
+ 
+TEST_P(ModelParameterizedTest, EfficientnNetTest)
+{
+    auto model_path = string_format(MODEL_PATH_TEMPLATE, GetParam().name.c_str(), GetParam().name.c_str());
+    auto model = ClassificationModel::create_model(DATA_DIR + "/" + model_path);
+    
+    auto ov_model = model->getModel();
+
+    ov::serialize(ov_model, "tmp_model.xml");
```
Wovchena, line 59: We will need to revise how tests are organized. I see the same problem with introducing tests for otx model.
for now maybe you should move this to `test_sanity.cpp`. You introduce lots of code copied from `test_sanity.cpp` just to test `model->getModel();` and `ov::serialize()`.

model_api/cpp/models/src/model_base.cpp
```diff
@@ -60,6 +60,18 @@ ModelBase::ModelBase(std::shared_ptr<ov::Model>& model, const ov::AnyMap& config
...
 }
 
+void ModelBase::updateModelInfo() {
+    if (!model) {
+        std::runtime_error("The ov::Model object is not accessible");
+    }
+
+    if (!inputsLayouts.empty()) {
+        auto layouts = formatLayouts(inputsLayouts);
+        model->set_rt_info(layouts, "model_info", "layout");
```
Wovchena, line 70: Not sure if we need this because we don't extract this info from IR. Plus MO itself can set these values in IR (not in rt_info)

Wovchena (changes requested):

tests/cpp/precommit/test_model_config.cpp
AlexKoff88: That is a draft of the test. The plan is to fill this file with many tests.

model_api/cpp/models/src/image_model.cpp
```diff
@@ -115,6 +115,18 @@ ImageModel::ImageModel(std::shared_ptr<InferenceAdapter>& adapter)
...
 
+void ImageModel::updateModelInfo() {
+    ModelBase::updateModelInfo();
+
+    model->set_rt_info(useAutoResize, "model_info", "auto_resize");
+    model->set_rt_info(formatResizeMode(resizeMode), "model_info", "resize_type");
+
+    if (!labels.empty()) {
+        model->set_rt_info(labels, "model_info", "labels");  
+    }
```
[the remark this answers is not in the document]
AlexKoff88: If they are empty no need to save them, right?

model_api/cpp/models/src/model_base.cpp
AlexKoff88: I think it is required to control the layout of input data so that the model could adjust to it and do preprocessing correctly.

model_api/cpp/models/src/image_model.cpp
```diff
@@ -115,6 +115,18 @@ ImageModel::ImageModel(std::shared_ptr<InferenceAdapter>& adapter)
     }
 }
 
+void ImageModel::updateModelInfo() {
+    ModelBase::updateModelInfo();
+
+    model->set_rt_info(useAutoResize, "model_info", "auto_resize");
```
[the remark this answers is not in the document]
AlexKoff88: this is what we should rename to `embed_preprocessing` as in Python. `embeded_processing` sounds better to me since we deal in postprocessing in some cases.

PR #45, opened 2023-03-27
[DRAFT]: Preprocessing embeding in C++ API

AlexKoff88:
- Added resize for classification models
- Added all the rest of resize modes
- Added save/load working for Classification and SSD models


model_api/cpp/utils/src/image_utils.cpp
```diff
@@ -50,6 +57,266 @@ cv::Mat resizeImageExt(const cv::Mat& mat, int width, int height, RESIZE_MODE re
...
+        case RESIZE_CROP:
+            return [=](const Output<Node>& node) {
+                return cropResizeGraph(node, size, interpolationMode);
+            };
+            break;
+        default:
+            throw std::runtime_error("Unsupported resize mode: " + resizeMode);
+            break;
+    }
+}
```
Wovchena, line 322:
```suggestion
}

```

model_api/cpp/utils/src/image_utils.cpp
```diff
@@ -50,6 +57,266 @@ cv::Mat resizeImageExt(const cv::Mat& mat, int width, int height, RESIZE_MODE re
         }
         break;
     }
+    case NO_RESIZE:
```
Wovchena, line 60: Do we need this mode?

Wovchena (changes requested):

model_api/cpp/utils/src/image_utils.cpp
AlexKoff88: I think it could be useful to state explicitly that we don't embed any interpolation into the model. E.g. if the model has been reshaped to the target resolution and no extra resize is required.

PR #49, opened 2023-03-29
Enable multilables classification

Wovchena:


model_api/cpp/models/src/classification_model.cpp
```diff
@@ -107,6 +130,31 @@ std::unique_ptr<ClassificationModel> ClassificationModel::create_model(std::shar
...
+    const ov::Tensor& logitsTensor = infResult.outputsData.find(outputNames[0])->second;
+    const float* logitsPtr = logitsTensor.data<float>();
+
+    ClassificationResult* result = new ClassificationResult(infResult.frameId, infResult.metaData);
+    auto retVal = std::unique_ptr<ResultBase>(result);
+
+    result->topLabels.reserve(labels.size());
+    for (size_t i = 0; i < labels.size(); ++i) {
+        float score = sigmoid(logitsPtr[i]);
+        if (score > 0.5f) {
```
AlexKoff88, line 149: is this 0.5f supposed to be a configurable parameter?

AlexKoff88 (approved):

model_api/cpp/models/src/classification_model.cpp
Wovchena: Well, its up to us :) I took the implementation from https://github.com/openvinotoolkit/training_extensions/blob/6754fdaa7d5dd4056c1bc8aaf3a8aa3afe13bd3d/otx/algorithms/classification/adapters/openvino/model_wrappers/openvino_models.py#L188. Here, the model itself doesn't allow to configure it and the default arg value is always used

AlexKoff88: ok, let's do the same for now.

PR #50, opened 2023-04-03
Align Python processing embedding with C++ 

Wovchena:


model_api/cpp/models/src/detection_model_ssd.cpp
```diff
@@ -35,6 +35,26 @@ struct InputData;
 
 std::string ModelSSD::ModelType = "ssd";
 
+ModelSSD::ModelSSD(std::shared_ptr<ov::Model>& model, const ov::AnyMap& configuration)
+    : DetectionModel(model, configuration) {
+    auto object_size_iter = configuration.find("object_size");
+    if (object_size_iter == configuration.end()) {
+        if (model->has_rt_info("model_info", "object_size")) {
+            objectSize = stoull(model->get_rt_info<std::string>("model_info", "object_size"));
```
AlexKoff88, line 43: Do we really have to provide the ability to configure these parameters because they are tightly connected to the model architecture and derived from it?  My take is that we should not.

AlexKoff88: The idea is that SSD requires these parameters to do postprocessing. In case of adapter SSD does not have access to the model to initialize them, so it fetches them from the configuration provided by the adapter. Everything is transparent to the user. I would like to highlight that these two parameters are not "configurable" so to say they come from the model itself.

Wovchena: Ok. I will revert this

docs/model-configuration.md
```diff
@@ -6,11 +6,11 @@ The list features only model wrappers which intoduce new configuration values in
...
 ### `ImageModel` and its subclasses
-1. `mean_values`: List - Normalization values, which will be subtracted from image channels for image-input layer during preprocessing
-1. `scale_values`: List - Normalization values, which will divide the image channels for image-input layer
-1. `reverse_input_channels`: bool - Reverse the input channel order
+1. `mean_values`: List - normalization values, which will be subtracted from image channels for image-input layer during preprocessing
+1. `scale_values`: List - normalization values, which will divide the image channels for image-input layer
+1. `reverse_input_channels`: bool - reverse the input channel order
 1. `resize_type`: str - crop, standard, fit_to_window or fit_to_window_letterbox
-1. `embed_preprocessing`: bool - Whether to embed preprocessing into the model
+1. `embedded_preprocessing`: bool - flag that pre/postprocessing embedded
```
AlexKoff88, line 13:
```suggestion
1. `embedded_processing`: bool - flag that pre/postprocessing embedded
```

AlexKoff88:
LGTM, just a small comment.

AlexKoff88 (approved):

PR #52, opened 2023-04-26
Test OTX segmentation models

Wovchena:
I have a feeling that comparing just unique values for segmentation is a weak test. I will update it later to compare histograms

### DIFF
--- a/docs/model-configuration.md
+++ b/docs/model-configuration.md
@@ -57,6 +57,9 @@ The list features only model wrappers which intoduce new configuration values in
 #### `SegmentationModel` and its subclasses
 1. `labels`: List - list of class labels
 1. `path_to_labels`: str - path to file with labels. Overrides the labels, if they sets via 'labels' parameter
+1. `blur_strength`: int - blurring kernel size. -1 signals not to apply bluring and soft_threshold
+1. `soft_threshold`: float - probability threshold value for bounding box filtering. inf signals not to apply bluring and soft_threshold
+1. `return_soft_prediction`: bool - retern raw resized model prediction in addition to processed one
 ### `Bert` and its subclasses
 1. `vocab`: Dict - mapping from string token to int
 1. `input_names`: str - comma-separated names of input layers
--- a/model_api/cpp/models/src/segmentation_model.cpp
+++ b/model_api/cpp/models/src/segmentation_model.cpp
@@ -32,7 +32,79 @@
 #include "models/input_data.h"
 #include "models/results.h"
 
+namespace {
+cv::Mat create_hard_prediction_from_soft_prediction(const cv::Mat& soft_prediction, float soft_threshold, int blur_strength) {
+    cv::Mat soft_prediction_blurred;
+    cv::blur(soft_prediction, soft_prediction_blurred, cv::Size{blur_strength, blur_strength});
+    assert(soft_prediction_blurred.channels() > 1);
+    cv::Mat hard_prediction{cv::Size{soft_prediction_blurred.rows, soft_prediction_blurred.cols}, CV_8UC1};
+    for (int i = 0; i < soft_prediction_blurred.rows; ++i) {
+        for (int j = 0; j < soft_prediction_blurred.cols; ++j) {
+            float max_prob = -std::numeric_limits<float>::infinity();
+            uint8_t max_id = 0;
+            for (int c = 0; c < soft_prediction_blurred.channels(); ++c) {
+                float prob = ((float*)soft_prediction_blurred.data)[
+                    i * soft_prediction_blurred.cols * soft_prediction_blurred.channels()
+                    + j * soft_prediction_blurred.channels()
+                    + c];
+                if (prob > max_prob) {
+                    max_prob = prob;
+                    max_id = c;
+                }
+            }
+            hard_prediction.at<uint8_t>(i, j) = max_id;
+        }
+    }
+    return hard_prediction;
+}
+}
+
 std::string SegmentationModel::ModelType = "Segmentation";
+
+SegmentationModel::SegmentationModel(std::shared_ptr<ov::Model>& model, const ov::AnyMap& configuration)
+        : ImageModel(model, configuration) {
+    auto blur_strength_iter = configuration.find("blur_strength");
+    if (blur_strength_iter == configuration.end()) {
+        if (model->has_rt_info("model_info", "blur_strength")) {
+            blur_strength = stoi(model->get_rt_info<std::string>("model_info", "blur_strength"));
+        }
+    } else {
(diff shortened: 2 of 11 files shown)


docs/model-configuration.md
```diff
@@ -57,6 +57,9 @@ The list features only model wrappers which intoduce new configuration values in
 #### `SegmentationModel` and its subclasses
 1. `labels`: List - list of class labels
 1. `path_to_labels`: str - path to file with labels. Overrides the labels, if they sets via 'labels' parameter
+1. `blur_strength`: int - blurring kernel size. -1 signals not to apply bluring and soft_threshold
```
AlexKoff88, line 60: `bluring` -> `blurring` everywhere.
`retern` -> `return`
`signals not to apply` sounds not well. I would rephrase to, e.g. `-1 value means no blurring and ...`

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -32,8 +32,80 @@
 #include "models/input_data.h"
 #include "models/results.h"
 
+namespace {
+cv::Mat create_hard_prediction_from_soft_prediction(const cv::Mat& soft_prediction, float soft_threshold, int blur_strength) {
+    cv::Mat soft_prediction_blurred;
```
AlexKoff88, line 37: @sovrasov, please take a look as well.

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -32,8 +32,80 @@
...
+        if (model->has_rt_info("model_info", "soft_threshold")) {
+            soft_threshold = stof(model->get_rt_info<std::string>("model_info", "soft_threshold"));
+        }
+    } else {
+        soft_threshold = soft_threshold_iter->second.as<float>();
+    }
+    auto return_soft_prediction_iter = configuration.find("return_soft_prediction");
+    if (return_soft_prediction_iter == configuration.end()) {
+        if (model->has_rt_info("model_info", "return_soft_prediction")) {
+            return_soft_prediction = model->get_rt_info<std::string>("model_info", "return_soft_prediction") == "True";
```
AlexKoff88, line 85: There is a conversion to bool tape as well as far as I remember. So, no need to compare strings.

tests/python/accuracy/public_scope.json
```diff
@@ -6,7 +6,55 @@
...
+        ]
+    },
+    {
+        "name": "otx_models/Lite-hrnet-18.xml",
+        "type": "SegmentationModel",
+        "test_data": [
+            {
+                "image": "coco128/images/train2017/000000000074.jpg",
+                "reference": [
+                    "0 1 "
```
AlexKoff88, line 21: I agree that this is not enough and should be enhanced. Normally, we should take references from OTX, and save them here. Maybe @sovrasov can help us to get such references. It should not take a lot of time. It's just a processed output 
 for one sample.

docs/model-configuration.md
Wovchena: Applied

model_api/cpp/models/src/segmentation_model.cpp
AlexKoff88: Ok, got it. OpenVINO is very separate as usual here in terms of representing the things )

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -32,8 +32,80 @@
 #include "models/input_data.h"
 #include "models/results.h"
 
+namespace {
+cv::Mat create_hard_prediction_from_soft_prediction(const cv::Mat& soft_prediction, float soft_threshold, int blur_strength) {
```
sovrasov, line 36: Why soft_threshold argument is not used here? Perhaps, thresholding makes no sense in most cases, since we have argmax, but in python you keep it.

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -32,8 +32,80 @@
...
+    cv::Mat soft_prediction_blurred;
+    cv::blur(soft_prediction, soft_prediction_blurred, cv::Size{blur_strength, blur_strength});
+    assert(soft_prediction_blurred.channels() > 1);
+    cv::Mat hard_prediction{cv::Size{soft_prediction_blurred.rows, soft_prediction_blurred.cols}, CV_8UC1};
+    for (int i = 0; i < soft_prediction_blurred.rows; ++i) {
+        for (int j = 0; j < soft_prediction_blurred.cols; ++j) {
+            float max_prob = -std::numeric_limits<float>::infinity();
+            uint8_t max_id = 0;
+            for (int c = 0; c < soft_prediction_blurred.channels(); ++c) {
+                float prob = ((float*)soft_prediction_blurred.data)[
```
sovrasov, line 46: Here Mat.ptr(i) could be used to reduce the indexing expression

tests/python/accuracy/public_scope.json
sovrasov: As soon as MAPI is a standalone tool, I'd propose not to tie to references from OTX. Instead, we may check C++ vs python output. Inconsistencies with OTX could be fixed once during the integration and later validation pytorch task vs openvino task will keep the OV wrappers in shape.

model_api/cpp/models/src/segmentation_model.cpp
Wovchena: Yes, that's strange. It makes sense to return the most probable label, not the one with the smallest index if every labels has low probability for a given pixel. Can I completely drop soft_prediction? But this will introduce difference between OTX and ModelAPI

model_api/cpp/models/src/segmentation_model.cpp
Wovchena: Updated

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -32,8 +32,80 @@
...
+            float max_prob = -std::numeric_limits<float>::infinity();
+            uint8_t max_id = 0;
+            for (int c = 0; c < soft_prediction_blurred.channels(); ++c) {
+                float prob = ((float*)soft_prediction_blurred.ptr(i, j))[c];
+                if (prob > max_prob) {
+                    max_prob = prob;
+                    max_id = c;
+                }
+            }
+            hard_prediction.at<uint8_t>(i, j) = max_id;
```
sovrasov, line 52: The result of discussion: set hard prediction to 0 if max_prob < soft_threshold

Wovchena: Updated

AlexKoff88 (approved):

AlexKoff88:
Let's merge.

PR #56, opened 2023-05-10
Prohibit whitespace errors and non ASCII chars

Wovchena:


.github/workflows/test_precommit.yml
```diff
@@ -164,3 +164,14 @@ jobs:
       run: |
         set PATH=opencv\opencv\build\x64\vc16\bin;w_openvino_toolkit_windows_2022.3.0.9052.9752fafe8eb_x86_64\runtime\bin\intel64\Release;w_openvino_toolkit_windows_2022.3.0.9052.9752fafe8eb_x86_64\runtime\3rdparty\tbb\bin;%PATH%
         .\build\Release\test_sanity.exe -d data -p tests\cpp\precommit\public_scope.json && .\build\Release\test_model_config -d data
+  Prohibit-whitespace-errors-and-non-ASCII-chars:
+    name: Prohibit whitespace errors and non ASCII chars
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
AlexKoff88, line 171: I think it should be a part of Code quality checks to avoid extra checkout and speedup the pre-commit. @Wovchena, what do you think? any reason to keep it as a separate job?

Wovchena: Moved. There's no specific reason

PR #57, opened 2023-05-10
Segmentation CPP generate annotations

RHeckerIntel:
# Implemented getting contour from segmentation

In order to do this I suggest a small rewrite from postprocessing.
Current implementation has some duplication and it does not allow soft prediction to be returned when the soft_threshold with blurring is disabled (soft_threshold = inf || blur_strength = -1).

Secondly I fixed a couple of issues with row/minor major and enable the segmentation test in ModelParameterize tests.


model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -217,41 +259,43 @@ void SegmentationModel::prepareInputsOutputs(std::shared_ptr<ov::Model>& model)
...
+        result->annotations = annotations;
+        result->resultImage = hard_prediction;
+        result->soft_prediction = soft_prediction;
+        return std::unique_ptr<ResultBase>(result);
+    } else {
+        ImageResult* result = new ImageResult(infResult.frameId, infResult.metaData);
+        result->annotations = annotations;
+        result->resultImage = cv::Mat(outHeight, outWidth, CV_8UC1);
 
         cv::resize(result->resultImage,
```
Wovchena, line 300: You just resize uninitialized cv::Mat here because you have `result->resultImage = cv::Mat(outHeight, outWidth, CV_8UC1);`

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -54,6 +68,34 @@ cv::Mat create_hard_prediction_from_soft_prediction(const cv::Mat& soft_predicti
     }
     return hard_prediction;
 }
+
+std::vector<Annotation> create_annotation_from_segmentation_map(const cv::Mat& hard_prediction, const cv::Mat& soft_prediction, std::vector<std::string> label_map) {
+    std::vector<Annotation> annotations = {};
+    cv::Mat label_index_map;
+    cv::Mat current_label_soft_prediction;
+    size_t labelID = 0;
+    for (int index = 1; index < soft_prediction.channels(); index++) {
```
Wovchena, line 77: If `1 == soft_prediction.channels()`, empty vector is returned. Is this intentional?

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -54,6 +68,34 @@ cv::Mat create_hard_prediction_from_soft_prediction(const cv::Mat& soft_predicti
...
+        cv::extractChannel(soft_prediction, current_label_soft_prediction, index);
+        cv::inRange(hard_prediction, cv::Scalar(index, index, index), cv::Scalar(index, index, index), label_index_map);
+        std::vector<std::vector<cv::Point>> contours;
+        cv::findContours(label_index_map, contours, cv::RETR_EXTERNAL, cv::CHAIN_APPROX_NONE);
```
Wovchena, line 81: Why is the contour based on `hard_prediction`, but `probability` is based on `soft_prediction`, should they be the same?

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -54,6 +68,34 @@ cv::Mat create_hard_prediction_from_soft_prediction(const cv::Mat& soft_predicti
     }
     return hard_prediction;
 }
+
+std::vector<Annotation> create_annotation_from_segmentation_map(const cv::Mat& hard_prediction, const cv::Mat& soft_prediction, std::vector<std::string> label_map) {
```
Wovchena, line 72:
```suggestion
std::vector<Annotation> create_annotation_from_segmentation_map(const cv::Mat& hard_prediction, const cv::Mat& soft_prediction, const std::vector<std::string>& label_map) {
```

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -54,6 +68,34 @@ cv::Mat create_hard_prediction_from_soft_prediction(const cv::Mat& soft_predicti
...
+    size_t labelID = 0;
+    for (int index = 1; index < soft_prediction.channels(); index++) {
+        cv::extractChannel(soft_prediction, current_label_soft_prediction, index);
+        cv::inRange(hard_prediction, cv::Scalar(index, index, index), cv::Scalar(index, index, index), label_index_map);
+        std::vector<std::vector<cv::Point>> contours;
+        cv::findContours(label_index_map, contours, cv::RETR_EXTERNAL, cv::CHAIN_APPROX_NONE);
+        for (auto &contour: contours) {
+            cv::Mat mask = cv::Mat::zeros(hard_prediction.rows, hard_prediction.cols, hard_prediction.type());
+            std::vector<std::vector<cv::Point>> localContours = { contour };
+            cv::drawContours(mask, localContours, -1, 255, -1);
```
Wovchena, line 85: `cv::drawContours()`'s `contourIdx` arg was probably introduced to remove the need to make this copy: `std::vector<std::vector<cv::Point>> localContours = { contour };`. I suggest using that feature.

model_api/cpp/models/include/models/results.h
```diff
@@ -122,6 +122,13 @@ struct RetinaFaceDetectionResult : public DetectionResult {
     std::vector<cv::Point2f> landmarks;
 };
 
+struct Annotation {
+    size_t labelID;
```
Wovchena, line 126: > I've temporarily used an index as labelID. This needs to be changed, but i'm unsure to what.

Why is `labelID` introduced in first place?

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -54,6 +68,34 @@ cv::Mat create_hard_prediction_from_soft_prediction(const cv::Mat& soft_predicti
...
+            float probability = cv::mean(current_label_soft_prediction, mask)[0];
+            annotations.push_back( {
+                    labelID,
+                    label_map[index],
```
Wovchena, line 89: You must handle the case when `label_map.size() >= index`. See https://github.com/openvinotoolkit/model_api/blob/e86a632d2a43bfa14d326c431023a1979f693b56/model_api/cpp/models/include/models/detection_model.h#L46

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -34,17 +34,31 @@
...
+
     assert(soft_prediction_blurred.channels() > 1);
-    cv::Mat hard_prediction{cv::Size{soft_prediction_blurred.rows, soft_prediction_blurred.cols}, CV_8UC1};
+    cv::Mat hard_prediction{cv::Size{soft_prediction_blurred.cols, soft_prediction_blurred.rows}, CV_8UC1};
     for (int i = 0; i < soft_prediction_blurred.rows; ++i) {
         for (int j = 0; j < soft_prediction_blurred.cols; ++j) {
             float max_prob = -std::numeric_limits<float>::infinity();
+            if (applySoftThreshold) {
+                max_prob = soft_threshold;
+            }
```
Wovchena, line 56: I suggest setting the default value of soft_threshold to -inf. This will allow to replace
```cpp
            float max_prob = -std::numeric_limits<float>::infinity();
            if (applySoftThreshold) {
                max_prob = soft_threshold;
            }
 ```
with just `float max_prob = soft_threshold`

**Review by Wovchena — CHANGES_REQUESTED** [review]
> I'm missing a public_scope.json there (or should we use the python one?

Use `public_scope.json` from `python`: https://github.com/openvinotoolkit/model_api/blob/e86a632d2a43bfa14d326c431023a1979f693b56/.github/workflows/test_accuracy.yml#L56

I've enabled precommit. It failed. Fix it please

**Review by Wovchena — CHANGES_REQUESTED** [review]
I forgot to ask the main question. What are these changes for?

**Comment by RHeckerIntel** [comment]
Thank you very much for the review. The main reason for the changes is that we want as much preprocessing to be done inside model_api. My goal is to be able to easily include it inside our mediapipe. Whilst it is relatively easy to implement these changes as a separate calculator I believe this is useful as a general solution.

I can imagine that you'd want this behind a configuration to save performance. But I'll leave that up to you.

**Comment by Wovchena** [comment]
> Thank you very much for the review. The main reason for the changes is that we want as much preprocessing to be done inside model_api. My goal is to be able to easily include it inside our mediapipe. Whilst it is relatively easy to implement these changes as a separate calculator I believe this is useful as a general solution.
> 
> I can imagine that you'd want this behind a configuration to save performance. But I'll leave that up to you.

How about having `create_annotation_from_segmentation_map()` as a stand alone function. So model wrapper produces outputs without contours, but if someone needs them, they can call `create_annotation_from_segmentation_map()` on the model output?

**Comment by Wovchena** [comment]
There are 6 hidden conversations which you didn't react anyhow to

File: model_api/cpp/models/src/segmentation_model.cpp [hunk 9]
```diff
@@ -34,17 +34,31 @@
 
 namespace {
 cv::Mat create_hard_prediction_from_soft_prediction(const cv::Mat& soft_prediction, float soft_threshold, int blur_strength) {
-    cv::Mat soft_prediction_blurred;
-    cv::blur(soft_prediction, soft_prediction_blurred, cv::Size{blur_strength, blur_strength});
+    cv::Mat soft_prediction_blurred = soft_prediction.clone();
+    if (soft_prediction.channels() == 1) {
+        return soft_prediction_blurred;
+    }
+    cv::Mat current_label_soft_prediction;
```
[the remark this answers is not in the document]
RHeckerIntel: Removed. Thanks.

model_api/cpp/models/include/models/results.h
```diff
@@ -122,6 +122,13 @@ struct RetinaFaceDetectionResult : public DetectionResult {
     std::vector<cv::Point2f> landmarks;
 };
 
+struct Annotation {
```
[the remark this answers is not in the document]
RHeckerIntel: I can see where you're coming from. But contour is used already in OpenCV and I'm adding more information like label and probability. Nevertheless I've renamed it to Contour in my latest commit. WDYT?

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -34,17 +34,31 @@
...
+
+    if (blur_strength > -1) {
```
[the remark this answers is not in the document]
RHeckerIntel: Ah yes. I missed this specification. Reverted back to the docs documentation.

model_api/cpp/models/src/segmentation_model.cpp
RHeckerIntel: Yes, this also caused the error in the test. I've moved the getLabelName to ImageModel and the get_annotations to segmentation_model class so that I can reuse it.

model_api/cpp/models/src/segmentation_model.cpp
RHeckerIntel: > How about having `create_annotation_from_segmentation_map()` as a stand alone function. So model wrapper produces outputs without contours, but if someone needs them, they can call `create_annotation_from_segmentation_map()` on the model output?

I do like the idea, but the returned object is a result object. The soft prediction is already lost at that point, which is required to generate the contours. It would make more sense to lock it behind a config in that case. WDYT?

Wovchena: There's `return_soft_prediction` that makes the soft prediction available. The function generating contours should try to cast the returned object to `ImageResultWithSoftPrediction` and simply throw if the cast fails.

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -217,41 +259,43 @@ void SegmentationModel::prepareInputsOutputs(std::shared_ptr<ov::Model>& model)
...
+        const auto data = outTensor.data<int64_t>();
+        for (size_t i = 0; i < predictions.total(); ++i) {
+            reinterpret_cast<int32_t*>(predictions.data)[i] = int32_t(data[i]);
+        }
+        predictions.convertTo(soft_prediction, CV_8UC1);
+    } else if (outTensor.get_element_type() == ov::element::f32) {
+        float* data = outTensor.data<float>();
+        std::vector<cv::Mat> channels;
+        for (size_t c = 0; c < outTensor.get_shape()[1]; ++c) {
+            channels.emplace_back(cv::Size{outHeight, outWidth}, CV_32F, data + c * outHeight * outWidth);
```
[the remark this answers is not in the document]
RHeckerIntel: cv::Size is {width, height} not the other way around.

model_api/cpp/models/include/models/results.h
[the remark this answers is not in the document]
Wovchena: You can name it `LabeledContour` then :)

model_api/cpp/models/src/segmentation_model.cpp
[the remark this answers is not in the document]
Wovchena: I'm starting to think that that was a good idea to split `blur_strength` and `soft_threshold`, so I would prefer to update docs instead. But no problem, I can do it myself later.

Wovchena:
> > There are 6 hidden conversations which you didn't react anyhow to
> 
> I agreed and applied the feedback.

Some of the conversions hidden by GitHub UI are still applicable. For example https://github.com/openvinotoolkit/model_api/pull/57#discussion_r1189979625

model_api/cpp/models/include/models/results.h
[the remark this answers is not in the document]
RHeckerIntel: Since I'm no longer returning by default, contour will do.

model_api/cpp/models/src/segmentation_model.cpp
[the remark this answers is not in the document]
RHeckerIntel: Done :)

model_api/cpp/models/src/segmentation_model.cpp
RHeckerIntel: Correct, first label is a background and not relevant for the contours.

model_api/cpp/models/src/segmentation_model.cpp
RHeckerIntel: Was implemented this way by OTX. Seems that it can just use soft_prediction data. Will simplify thanks

model_api/cpp/models/src/segmentation_model.cpp
RHeckerIntel: Done.

model_api/cpp/models/src/segmentation_model.cpp
RHeckerIntel: Fixed.

RHeckerIntel:
Well that was a messy PR. Sorry, there was quite a lot to figure out and learn. Thank you for the patience.

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -204,73 +212,70 @@ std::unique_ptr<ResultBase> SegmentationModel::postprocess(InferenceResult& infR
...
-        channels.emplace_back(cv::Size{outHeight, outWidth}, CV_32F, data + c * outHeight * outWidth);
+        predictions.convertTo(soft_prediction, CV_8UC1);
+    } else if (outTensor.get_element_type() == ov::element::f32) {
+        float* data = outTensor.data<float>();
+        std::vector<cv::Mat> channels;
+        for (size_t c = 0; c < outTensor.get_shape()[1]; ++c) {
+            channels.emplace_back(cv::Size{outWidth, outHeight}, CV_32FC1, data + c * outHeight * outWidth);
+        }
+        cv::merge(channels, soft_prediction);
+        cv::resize(soft_prediction, soft_prediction, {inputImgSize.inputImgWidth, inputImgSize.inputImgHeight}, 0.0, 0.0, cv::INTER_NEAREST);
```
Wovchena, line 233: You've changed the algo. Originally it was:
1. blur
2. resize

now it is
1. resize
2. blur

Is this intentional?

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -204,73 +212,70 @@ std::unique_ptr<ResultBase> SegmentationModel::postprocess(InferenceResult& infR
...
-    cv::merge(channels, soft_prediction);
-    cv::Mat hard_prediction = create_hard_prediction_from_soft_prediction(soft_prediction, soft_threshold, blur_strength);
-    cv::resize(hard_prediction, hard_prediction, {inputImgSize.inputImgWidth, inputImgSize.inputImgHeight}, 0.0, 0.0, cv::INTER_NEAREST);
+    cv::Mat hard_prediction = create_hard_prediction_from_soft_prediction(
+        soft_prediction, soft_threshold, blur_strength);
+
     if (return_soft_prediction) {
-        cv::resize(soft_prediction, soft_prediction, {inputImgSize.inputImgWidth, inputImgSize.inputImgHeight}, 0.0, 0.0, cv::INTER_NEAREST);
         ImageResultWithSoftPrediction* result = new ImageResultWithSoftPrediction(infResult.frameId, infResult.metaData);
         result->resultImage = hard_prediction;
```
Wovchena, line 240: `hard_prediction` may or may not be resized:
else branch of `if (return_soft_prediction)` always resizes but if the condition is true, the result depends on whether `soft_prediction` was resized which depends on `outTensor.get_element_type()`

Make it consistent

Wovchena (changes requested):

model_api/cpp/models/src/segmentation_model.cpp
Wovchena: Please throw if someone calls `getContours()` and `imageResult->soft_prediction.channels()` is 1

model_api/cpp/models/src/segmentation_model.cpp
Wovchena: I think you still use both `hard_prediction` and `soft_prediction`

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -169,21 +182,16 @@ void SegmentationModel::prepareInputsOutputs(std::shared_ptr<ov::Model>& model)
 
     const ov::Layout& inputLayout = getInputLayout(input);
     const ov::Shape& inputShape = input.get_partial_shape().get_max_shape();
-    if (inputShape.size() != 4 || inputShape[ov::layout::channels_idx(inputLayout)] != 3) {
+    if (inputShape.size() != 4 ||
+        inputShape[ov::layout::channels_idx(inputLayout)] != 3) {
```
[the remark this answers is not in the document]
RHeckerIntel: I have a clang-format. Can we use  one so that we can skip these issues?

model_api/python/openvino/model_api/models/segmentation.py
```diff
@@ -104,50 +109,66 @@ def parameters(cls):
...
+                current_label_soft_prediction = soft_prediction
+
+            obj_group = img_class == label_index
+            label_index_map = (obj_group.T.astype(int) * 255).astype(np.uint8)
+
+            # Contour retrieval mode CCOMP (Connected components) creates a two-level
+            # hierarchy of contours
+            contours, _hierarchy = cv2.findContours(
+                label_index_map, cv2.RETR_EXTERNAL, cv2.CHAIN_APPROX_NONE
+            )
```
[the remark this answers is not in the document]
RHeckerIntel: Did not implement probability in the python version. Will do.

model_api/cpp/models/include/models/detection_model.h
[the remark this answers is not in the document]
RHeckerIntel: The existing model checks should suffice no? I just refactored the postprocessing code a bit since it had a bit of duplication and bugs.

model_api/python/openvino/model_api/models/segmentation.py
```diff
@@ -104,50 +109,66 @@ def parameters(cls):
...
-                interpolation=cv2.INTER_NEAREST,
-            )
             return hard_prediction, soft_prediction
         return hard_prediction
 
+    def get_contours(
+        self, hard_prediction: np.ndarray, soft_prediction: np.ndarray
+    ) -> list:
+        height, width = hard_prediction.shape[:2]
+        img_class = hard_prediction.swapaxes(0, 1)
```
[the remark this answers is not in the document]
RHeckerIntel: `obj_group = img_class == label_index`  is used to get the mask for that specific label.

model_api/cpp/models/src/segmentation_model.cpp
RHeckerIntel: Thanks! good catch.

model_api/cpp/models/src/segmentation_model.cpp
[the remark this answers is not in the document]
Wovchena: It seems clang-format is applied only partially to the file. Probably only the lines which you modified. This results in file being inconsistent with itself.

Applying clang-format to the whole file also is not a solution. That would blow up PR's diff.

My suggestion is to revert the changes of lines which you didn't actually modify. And after that PR, if you wish you can add checks to enforce a coding style. That should be OpenVINO's style https://github.com/openvinotoolkit/openvino/blob/master/docs/dev/coding_style.md

model_api/python/openvino/model_api/models/segmentation.py
```diff
@@ -41,10 +41,15 @@ def create_hard_prediction_from_soft_prediction(
     Returns:
         Numpy array of the hard prediction
     """
-    soft_prediction_blurred = cv2.blur(soft_prediction, (blur_strength, blur_strength))
-    assert len(soft_prediction.shape) == 3
-    soft_prediction_blurred[soft_prediction_blurred < soft_threshold] = 0
-    return np.argmax(soft_prediction_blurred, axis=2)
+    if blur_strength == -1 and soft_threshold == float("inf"):
```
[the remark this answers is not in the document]
RHeckerIntel: Agree, but still the soft_threshold behavior should be the same as per our docs: ```soft_threshold: float - probability threshold value for bounding box filtering. **inf** value means no blurring and no soft_threshold```

model_api/cpp/models/include/models/detection_model.h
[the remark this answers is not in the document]
Wovchena: Yes, my bad. But new `getContours()` (and python's equivalent) isn't tested now. Will you add this call into tests so they dumped the results of `getContours()` and compared between C++ and Python.

model_api/cpp/models/src/segmentation_model.cpp
Wovchena: I agree that resize must be applied sooner or later. The question is only about the order

model_api/python/openvino/model_api/models/segmentation.py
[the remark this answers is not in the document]
Wovchena: Yes, sorry

model_api/cpp/models/src/segmentation_model.cpp
RHeckerIntel: Got confirmation that it's blur first, resize later. Will apply.

model_api/cpp/models/src/segmentation_model.cpp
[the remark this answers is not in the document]
RHeckerIntel: Agree completely.

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -28,24 +28,35 @@
...
-    cv::Mat hard_prediction{cv::Size{soft_prediction_blurred.rows, soft_prediction_blurred.cols}, CV_8UC1};
+    cv::Mat soft_prediction_blurred = soft_prediction.clone();
+    if (soft_prediction.channels() == 1) {
+        return soft_prediction_blurred;
+    }
+
+    bool applyBlurAndSoftThreshold = (blur_strength > -1 && soft_threshold < std::numeric_limits<float>::infinity());
+    if (applyBlurAndSoftThreshold) {
+        cv::blur(soft_prediction_blurred, soft_prediction_blurred, cv::Size{blur_strength, blur_strength});
+    }
```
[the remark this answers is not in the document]
Wovchena: Actually no.  I changed my mind :) Why would someone hwo need just hard oredictions pay for that copy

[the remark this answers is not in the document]
RHeckerIntel: Hah, fair enough. Good point.

model_api/python/openvino/model_api/models/segmentation.py
```diff
@@ -41,10 +41,15 @@ def create_hard_prediction_from_soft_prediction(
     Returns:
         Numpy array of the hard prediction
     """
-    soft_prediction_blurred = cv2.blur(soft_prediction, (blur_strength, blur_strength))
-    assert len(soft_prediction.shape) == 3
-    soft_prediction_blurred[soft_prediction_blurred < soft_threshold] = 0
-    return np.argmax(soft_prediction_blurred, axis=2)
+    if blur_strength == -1 or soft_threshold == float("inf"):
```
RHeckerIntel, line 44: @Wovchena In the original code this if statement was flipped from what I can see. We discussed this in the past but the master implementation in c++ and python both use if (blur_strength == -1 and soft_threshold == infinity) else blur. But that should be an or based on the documentation. If either is soft_threshold or blur_strength are infinity or -1 (respectively) it should *not* blur.

I've applied this in this PR.

model_api/python/openvino/model_api/models/segmentation.py
```diff
@@ -104,50 +109,66 @@ def parameters(cls):
...
 
+    def get_contours(
+        self, hard_prediction: np.ndarray, soft_prediction: np.ndarray
+    ) -> list:
+        height, width = hard_prediction.shape[:2]
+        img_class = hard_prediction.swapaxes(0, 1)
+
+        # pylint: disable=too-many-nested-blocks
+        combined_contours = []
+        for label_index, label in enumerate(self.labels):
```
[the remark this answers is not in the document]
RHeckerIntel: Done.

Wovchena:
FYI https://github.com/openvinotoolkit/model_api/pull/67#issuecomment-1570481771

Wovchena:
https://github.com/openvinotoolkit/model_api/pull/72 is merged. Please, resolve merge conflicts.

tests/python/accuracy/public_scope.json
```diff
@@ -1,12 +1,26 @@
 [
-    {
+     {
```
Wovchena, line 2:
```suggestion
    {
```

model_api/cpp/models/src/segmentation_model.cpp
```diff
@@ -208,73 +222,70 @@ std::unique_ptr<ResultBase> SegmentationModel::postprocess(InferenceResult& infR
...
         result->soft_prediction = soft_prediction;
         return std::unique_ptr<ResultBase>(result);
+    } else {
+        ImageResult* result = new ImageResult(infResult.frameId, infResult.metaData);
+        result->resultImage = hard_prediction;
+        return std::unique_ptr<ResultBase>(result);
     }
-    ImageResult* result = new ImageResult(infResult.frameId, infResult.metaData);
-    result->resultImage = hard_prediction;
-    return std::unique_ptr<ResultBase>(result);
```
Wovchena, line 274: You don't need `else` if `if` ends with `return`. Revert this please

model_api/cpp/models/src/segmentation_model.cpp
Wovchena: This is still applicable

model_api/cpp/models/src/segmentation_model.cpp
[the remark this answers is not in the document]
Wovchena: You should revert the changes of lines which you didn't actually modify as we agreed

model_api/python/openvino/model_api/models/segmentation.py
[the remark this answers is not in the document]
Wovchena: You shouldn't return empty list immediately `if self.labels is None`. It still possible to extract contours. They are just won't have string representation

Wovchena (changes requested):

RHeckerIntel:
What is applicable?

Wovchena (changes requested):
> What is applicable?

GitHub likes to combine all review comments into one batch. Given that, for an already started conversations GitHub puts two copies of a comment. One is in that batch (which you noticed) and one under the actual conversation. You need to find the comments under the actual conversations. The way I usually deal with that is the following: view all hidden conversations by clicking `Load more...` and search for the occurrences of `This is still applicable`.

model_api/cpp/models/src/segmentation_model.cpp
Wovchena: I didn't know that there are people who prefer that style.

The problem with changing style in scope of other work is that a reviewer may not like the change or simply be unaware of other style. In that context such changes cause questions.

I'm not saying that you shouldn't modify style in a PR at all. That's very convenient to fix something minor or stylish while you have a file opened. I'm saying that, that the bigger piece of code you modify the more reasons a reviewer has to ask questions/start an argument.

RHeckerIntel:
I added the throw error when soft_prediction only has one layer.
I however left the usage of hard_prediction in get_contours. Using only soft prediction is possible, but it would generally mean creating the hard_prediction again. It makes more sense to use the already calculated version.

Wovchena (approved):

PR #66, opened 2023-05-16
Add rotated rects

Wovchena:


tests/python/accuracy/test_accuracy.py
```diff
@@ -82,6 +83,8 @@ def test_image_models(data, dump, result, model_data):
         outputs = model(image)
         if not isinstance(outputs, list):
             outputs = [outputs]
+        if model_data["type"] == MaskRCNNModel.__name__:
+            outputs = add_rotated_rects(outputs)
```
sovrasov, line 87: Is it valid for all MRCNN models? We have ordinary IS models as well

Wovchena: I don't think so. You've shared only OTX `MaskRCNNModel` models so far. So only they are verifyed

sovrasov: That's true, but these models are semantic segmentation ones and they are not supposed to output rotated rects

Wovchena: `add_rotated_rects()` is intentionally made a to be a separate function. If one doesn't want rotated rects, they don't call it. This test implementation choose is to always call it because it is applicable for every `MaskRCNNModel` in tests. Even if such models that don't require rects are added to tests, the test is still free to call `add_rotated_rects()` because it won't harm.

PR #76, opened 2023-06-02
Add raw_scores for cls models

sovrasov:
# What does this PR do?

This PR is just to start a discussion. In OTX these raw scores are required, and perhaps they are also useful for advanced users, who want to analyze all cls probs.

Fixes # (issue)

## Before submitting


Wovchena:
Well, OK. You just need to allow cases when there's no `raw_scores`

model_api/cpp/models/src/classification_model.cpp
```diff
@@ -109,16 +109,30 @@ void addOrFindSoftmaxAndTopkOutputs(std::shared_ptr<ov::Model>& model, size_t to
 
     auto indices = std::make_shared<ov::op::v0::Result>(topkNode->output(0));
     auto scores = std::make_shared<ov::op::v0::Result>(topkNode->output(1));
-    model = std::make_shared<ov::Model>(ov::ResultVector{scores, indices}, model->get_parameters(), "classification");
+    ov::ResultVector results_vector;
+    if (add_raw_scores) {
+        auto raw_scores = std::make_shared<ov::op::v0::Result>(softmaxNode->output(0));
```
Wovchena, line 114: I just realized, setting `topk` to the actual number of classes will give you the equivalent of `raw_scores`, but also sorted. So by introducing `raw_scores` you only save time on sorting. I'm ok, if you need this. I just wanted to highlight this.

Wovchena: One more thing. topk doesn't actually sort. So the overhead is only on copying scores from a tensor to C++ struct

sovrasov: Is it guaranteed that topk doesn't change toe order of output scores?

Wovchena: @sovrasov, why does it matter?

Docstrings in a header don't specify this. But I think this a reasonable expectation. @tomdol, can you comment, if `TopK` may change the order of output if `k` is equal to the length of `axis`?

sovrasov: If the order is twisted, it is impossible to establish label -> score correspondence in OTX

Wovchena: Can you name, what kind of applied postprocessing is not desired for you?

sovrasov: I need to get raw cls scores after softmax without any changes in order, like it's done for multilabel and hierarchical classification. Any kind of shrinking the amount of the cls scores (like now, to reducing 1) or reordering shouldn't be applied to raw_scores.

sovrasov: As far as I understand, you propose changing k to number of classes in topk and then manually select the highest score in postprocess, but I'd remove topk then.

Wovchena: Not really. The idea is to configure the model to set topk. That way allows you to get all scores using big topk, while keeping small topk for other use cases.

As for the scores reorder, technically it is possible to restore the order after (if) topk shuffles them because you know the order of labels, and labels are given in final postprocess() structure.

Wovchena: > second topk

There's only one topk. You can assign a value to it.

> if I need to get only scores and not going to use that scores in postprocess?

scores aren't used much in postprocess anyway. They are simply copied to the output struct.

PR #84, opened 2023-06-10
Update workflows

Wovchena:
1. Remove running on `schedule`: don't generate extra entry in `Actions` tab of the repo, we run all checks on `pull_request` anyway
1. Run checks for all `pull_request`, not only for `master`
1. Use `cache: pip`
1. Enable macOS test


.github/workflows/test_precommit.yml
```diff
@@ -171,24 +166,41 @@ jobs:
...
-      strategy:
-        fail-fast: false
-        matrix:
-          python-version: [3.7, 3.8, 3.9, '3.10', '3.11']
-          os: [ubuntu-20.04, ubuntu-latest]
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04, ubuntu-22.04, macos-11, macos-12]
+        python-version: [3.7, 3.8, 3.9, '3.10', '3.11']
```
AlexKoff88, line 173: too many versions of Python. The normal approach is to test only 3.7 and 3.11. what do you think?

AlexKoff88 (approved):

.github/workflows/test_precommit.yml
Wovchena: I'd say it should also include the default python version for a given OS in addition to lower and upper bounds.

Extending that list costs me only one entry in matrix, so I thought why not to extend it. If I notice that there are queues that delay overall build results, I will shrink the list.

AlexKoff88: Ok, let it be. Now, it is not a bottleneck for CI.

PR #92, opened 2023-06-22
fix accuracy results

dtrawins:
# What does this PR do?
Fixing accuracy results with openvino adapter when sending multiple requests
Fixes # (issue)
Symptoms were with identical results returned for different sequential requests

## Before submitting

### DIFF
--- a/model_api/python/openvino/model_api/adapters/openvino_adapter.py
+++ b/model_api/python/openvino/model_api/adapters/openvino_adapter.py
@@ -298,7 +298,7 @@ class OpenvinoAdapter(InferenceAdapter):
 
     def infer_sync(self, dict_data):
         self.infer_request = self.async_queue[self.async_queue.get_idle_request_id()]
-        result= self.infer_request.infer(dict_data)
+        result = self.infer_request.infer(dict_data)
         return self.get_raw_result(result)
 
     def infer_async(self, dict_data, callback_data) -> None:
--- a/model_api/python/openvino/model_api/adapters/openvino_adapter.py
+++ b/model_api/python/openvino/model_api/adapters/openvino_adapter.py
@@ -287,7 +287,7 @@ class OpenvinoAdapter(InferenceAdapter):
         }
         self.model.reshape(new_shape)
 
-    def get_raw_result(self, request):
+    def get_raw_result(self, result):
         raw_result = {key.get_any_name(): data for key, data in result.items()}
         return raw_result
 
--- a/model_api/python/openvino/model_api/adapters/openvino_adapter.py
+++ b/model_api/python/openvino/model_api/adapters/openvino_adapter.py
@@ -288,7 +288,8 @@ class OpenvinoAdapter(InferenceAdapter):
         self.model.reshape(new_shape)
 
     def get_raw_result(self, request):
-        return {key: request.get_tensor(key).data for key in self.get_output_layers()}
+        raw_result = {key.get_any_name(): data for key, data in result.items()}
+        return raw_result
 
     def copy_raw_result(self, request):
         return {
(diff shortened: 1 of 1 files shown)


model_api/python/openvino/model_api/adapters/openvino_adapter.py
```diff
@@ -288,7 +288,8 @@ def reshape_model(self, new_shape):
         self.model.reshape(new_shape)
 
     def get_raw_result(self, request):
-        return {key: request.get_tensor(key).data for key in self.get_output_layers()}
+        raw_result = {key.get_any_name(): data for key, data in result.items()}
```
Wovchena, line 291: 'result' is not defined.

model_api/python/openvino/model_api/adapters/openvino_adapter.py
```diff
@@ -297,8 +298,8 @@ def copy_raw_result(self, request):
 
     def infer_sync(self, dict_data):
         self.infer_request = self.async_queue[self.async_queue.get_idle_request_id()]
-        self.infer_request.infer(dict_data)
-        return self.get_raw_result(self.infer_request)
+        result= self.infer_request.infer(dict_data)
+        return self.get_raw_result(result)
```
Wovchena, line 302: How does this solve
> Symptoms were with identical results returned for different sequential requests

? You simply moved from `get_tensor()` API to `infer()` return value

Wovchena (changes requested):

model_api/python/openvino/model_api/adapters/openvino_adapter.py
dtrawins: fixed

model_api/python/openvino/model_api/adapters/openvino_adapter.py
dtrawins: the main issue is in the infer_sync method which is returning the same results for sequential calls.
the fix was verified  in the demo from https://github.com/openvinotoolkit/model_server/tree/stable-diffusion/demos/stable-diffusion

Wovchena: I understand the problem statement. I don't understand how this patch solves it. Can you explain?

dtrawins: it makes a copy of the response so it is not reusing the previous response

Wovchena: You probably misuse API. Is `infer_sync()` used in multithreaded environment? There's `copy_raw_result()` which copies data and `infer_async()` for running multiple inferences at the same time.

dtrawins: I'm afraid the problem is even in a single threaded application. Sequential infer_sync calls give incorrect responses.
copy_raw_results is not used as it is not a part of the inference adapter and is not needed in ovms adapter.

Wovchena: I just realized. Since you are fine with this implementation, this means that `self.infer_request.infer(dict_data)` returns a copy as a result. And this copy is performed regardless of whether this implementation stores the returned value or discards.

The only possible problem is that model output tensor may have multiple names, but `result` dict may contain only one of them as key. For now I can't name a use case where this becomes an issue, so I think the problem may be ignored for now.

This means the overall idea of your PR is acceptable. Before I merge this, please, update `get_raw_result()` so the keys contained in returned value were from `self.get_output_layers()` but not just `get_any_name()`.

sovrasov:
Closed as outdated

PR #67, opened 2023-05-23
Add postprocessing for hierarchical classification

sovrasov:
# What does this PR do?

This PR add parsing of hierarchical classification output with further labels resolving.

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests?

PR #94, opened 2023-06-29
Reduce cmake_minimum version back to 3.10

RHeckerIntel:
Mediapipe and bazel (rules_foreign_cc) cannot handle that high of a cmake version. Mediapipe  uses rules_foreign_cc to build model_api and the highest cmake currently supported in rules_foreign_cc(0.9) is 3.23. Cmake 3.23 does not implement BUILD_LOCAL_INTERFACE used in model_api.

Can we reduce the cmake_minimum version back to 3.10?
And can we skip the target_link_libraries for json?

### DIFF
--- a/model_api/cpp/CMakeLists.txt
+++ b/model_api/cpp/CMakeLists.txt
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-cmake_minimum_required(VERSION 3.26)
+cmake_minimum_required(VERSION 3.10)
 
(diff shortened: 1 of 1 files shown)


Wovchena:
JSON is required. See https://github.com/openvinotoolkit/model_api/pull/67. I will come back to that after Tuesday. But maybe you can find a solution before that.

RHeckerIntel:
Ok. JSON required & `BUILD_LOCAL_INTERFACE` was best solution for windows support. Will close this and solve on mediapipe end.

PR #95, opened 2023-07-03
Add xai for MaskRCNNModel

Wovchena:


model_api/python/openvino/model_api/models/utils.py
```diff
@@ -88,17 +82,26 @@ def __init__(self, segmented_object):
...
 
-    def __str__(self):
-        return self.__to_str()
 
-    def __repr__(self):
-        return self.__to_str()
+class InstanceSegmentationResult(NamedTuple):
+    segmentedObjects: List[Union[SegmentedObject, SegmentedObjectWithRects]]
+    # Contan per class saliency_maps and "feature_vector" model output if feature_vector exists
+    saliency_map: List[np.ndarray]
```
sovrasov, line 95: In other Result objects saliency_map is just an np.ndarray. In OTX it's a list, but here we can unify the representation. Is there any particular reason to use list here?

Wovchena: Missing classes are represented not as  `np.ndarray` in that list. Should I fill the corresponding axes with zeros?

model_api/cpp/models/src/instance_segmentation.cpp
```diff
@@ -258,8 +328,16 @@ std::unique_ptr<ResultBase> MaskRCNNModel::postprocess(InferenceResult& infResul
         } else {
             obj.mask = raw_cls_mask;
         }
-        result->segmentedObjects.push_back(obj);
-
+        if (confidence > confidence_threshold) {
+            result->segmentedObjects.push_back(obj);
+        }
+        if (has_feature_vector_name) {
+            saliency_maps[obj.labelID - 1].push_back(obj.mask);
```
sovrasov, line 335: Is that mask resized, or we don't have a flag similar to postprocess_semantic_masks?

Wovchena: Resized. There's on top
```Cpp
        if (postprocess_semantic_masks) {
            obj.mask = segm_postprocess(obj, raw_cls_mask, internalData.inputImgHeight, internalData.inputImgWidth);
        }
```

sovrasov (approved):

PR #96, opened 2023-07-06
Add wrapper for anomaly model

ashwinvaidya17:
# Description
- Adds anomaly model wrapper.
- Should be merged after https://github.com/openvinotoolkit/training_extensions/pull/2305

## Sample predictions
(Visualization script attached. Rename `run.txt` to `run.py`)
image
[run.txt](https://github.com/openvinotoolkit/model_api/files/11967840/run.txt)


model_api/python/openvino/model_api/models/types.py
```diff
@@ -172,7 +172,7 @@ def __init__(
     def from_str(self, value):
         try:
             floats = [float(i) for i in value.split()]
-            ints = [int(i) for i in value.split()]
+            ints = [int(float(i)) for i in value.split()]
```
samet-akcay, line 175: not sure if I understand this? What's the motivation here?

ashwinvaidya17: When the string is something like this `"7.8 4.1 3.2"`, performing conversion like `int('7.8')` throws an exception. String to int conversion requires that the string does not contain decimal values.

samet-akcay: Thanks! I think this would be a nice comment in the code :D

samet-akcay (approved):
thanks looking good!

model_api/python/openvino/model_api/models/anomaly.py
```diff
@@ -0,0 +1,154 @@
...
+            pred_mask = cv2.resize(
+                pred_mask, (meta["original_shape"][1], meta["original_shape"][0])
+            )
+
+        if hasattr(self, "task") and self.task == "detection":
+            pred_boxes = self._get_boxes(pred_mask)
+            box_labels = np.ones(pred_boxes.shape[0])
+
+        return AnomalyResult(
+            anomaly_map=anomaly_map,
```
sovrasov, line 97: Does that task handles both detection and segmentation at once?

ashwinvaidya17: Yes. Since bounding boxes are computed from the mask, the code just checks if the task type is detection and adds bounding boxes to the results as well.

model_api/python/openvino/model_api/models/anomaly.py
```diff
@@ -0,0 +1,154 @@
...
+from .types import ListValue, NumericalValue, StringValue
+from .utils import AnomalyResult
+
+
+class AnomalyDetection(ImageModel):
+    __model__ = "AnomalyDetection"
+
+    def __init__(self, inference_adapter, configuration=None, preload=False):
+        super().__init__(inference_adapter, configuration, preload)
+        self.output_name = self.inference_adapter.model.outputs[0].any_name
```
sovrasov, line 36: self.outputs = self.inference_adapter.get_output_layers() already defined in Model, so you could access self.outputs here

model_api/python/openvino/model_api/models/utils.py
```diff
@@ -14,6 +14,8 @@
  limitations under the License.
 """
 
+from __future__ import annotations
```
sovrasov, line 17: Why do we need this?

model_api/python/openvino/model_api/models/anomaly.py
```diff
@@ -0,0 +1,154 @@
...
+
+        predictions = outputs[self.output_name]
+
+        if len(predictions.shape) == 1:
+            pred_score = predictions
+        else:
+            anomaly_map = predictions.squeeze()
+            pred_score = anomaly_map.reshape(-1).max()
+
+        if hasattr(self, "image_threshold"):
```
sovrasov, line 72: image_threshold attribute always exists. If it's absent in rt_info, the default values is assigned.

model_api/python/openvino/model_api/models/anomaly.py
```diff
@@ -0,0 +1,154 @@
...
+            pred_label = "Anomalous" if pred_score > self.image_threshold else "Normal"
+
+        if hasattr(self, "task") and self.task in ("segmentation", "detection"):
```
sovrasov, line 75: ditto

sovrasov:
@Wovchena please have a look from the overall mapi design perspective

model_api/python/openvino/model_api/models/utils.py
ashwinvaidya17: For python versions below `3.10` we can't use ` np.ndarray | None` notation without `__future__`. Otherwise we will have to use `Union[np.ndarray, None]`. This just looks cleaner 🙂

model_api/python/openvino/model_api/models/utils.py
```diff
@@ -22,6 +24,17 @@
 import numpy as np
 
 
+class AnomalyResult(NamedTuple):
+    """Results for anomaly models."""
+
+    anomaly_map: np.ndarray | None
+    box_labels: np.ndarray | None
+    pred_boxes: np.ndarray | None
```
samet-akcay, line 32: Is there any specific reason why we do not assign these to `None`

ashwinvaidya17: Yeah they should be None as well

model_api/python/openvino/model_api/models/anomaly.py
Wovchena, line 1: Will you add C++ implementation?

model_api/python/openvino/model_api/models/types.py
```diff
@@ -172,7 +172,9 @@ def __init__(
     def from_str(self, value):
         try:
             floats = [float(i) for i in value.split()]
-            ints = [int(i) for i in value.split()]
+            # In certain instances when a string contains a list of floats like "1.0 2.0 3.0",
+            # directly converting to ints will raise a ValueError
+            ints = [int(float(i)) for i in value.split()]
```
Wovchena, line 177: Here is even better idea. You can wrap `ints = [int(i) for i in value.split()]` with `try except ValueError`. With that you won't need to explain that the code may throw ValueError because, well, you explicitly state that with the implementation. And in comments you may still explain why you expectb `ValueError` to be risen.

Wovchena (changes requested):

model_api/python/openvino/model_api/models/anomaly.py
ashwinvaidya17: Yes! I am working on it atm.

model_api/python/openvino/model_api/models/anomaly.py
```diff
@@ -0,0 +1,152 @@
...
+            anomaly_map = cv2.resize(
+                anomaly_map, (meta["original_shape"][1], meta["original_shape"][0])
+            )
+            pred_mask = cv2.resize(
+                pred_mask, (meta["original_shape"][1], meta["original_shape"][0])
+            )
+
+        if self.task == "detection":
+            pred_boxes = self._get_boxes(pred_mask)
+            box_labels = np.ones(pred_boxes.shape[0])
```
Wovchena, line 92: `box_labels` is either `None` or `ones`. It doesn't provide any info. Are you sure you need it?

model_api/python/openvino/model_api/models/anomaly.py
```diff
@@ -0,0 +1,152 @@
...
+        return parameters
+
+    def _normalize(self, tensor: np.ndarray, threshold: float) -> np.ndarray:
+        """Currently supports only min-max normalization."""
+        normalized = ((tensor - threshold) / (self.max - self.min)) + 0.5
+        normalized = np.clip(normalized, 0, 1)
+        return normalized
+
+    @staticmethod
+    def _get_boxes(mask: np.ndarray) -> np.ndarray:
```
Wovchena, line 137: I suggest moving this to a standalone function. That way you won't pollute `AnomalyDetection`'s namespace and also this clarifies that `_get_boxes()` doesn't depend on `AnomalyDetection` class or its instance state.

model_api/python/openvino/model_api/models/anomaly.py
```diff
@@ -0,0 +1,152 @@
...
+from .image_model import ImageModel
+from .types import ListValue, NumericalValue, StringValue
+from .utils import AnomalyResult
+
+
+class AnomalyDetection(ImageModel):
+    __model__ = "AnomalyDetection"
+
+    def __init__(self, inference_adapter, configuration=None, preload=False):
+        super().__init__(inference_adapter, configuration, preload)
```
Wovchena, line 35:
```suggestion
        super().__init__(inference_adapter, configuration, preload)
        self._check_io_number(1, 1)
```

model_api/python/openvino/model_api/models/anomaly.py
```diff
@@ -0,0 +1,152 @@
...
+
+        if len(predictions.shape) == 1:
+            pred_score = predictions
+        else:
+            anomaly_map = predictions.squeeze()
+            pred_score = anomaly_map.reshape(-1).max()
+
+        pred_label = "Anomalous" if pred_score > self.image_threshold else "Normal"
+
+        if self.task in ("segmentation", "detection"):
```
Wovchena, line 73: What are the other values that `self.task` may have?

model_api/python/openvino/model_api/models/anomaly.py
```diff
@@ -0,0 +1,152 @@
...
+                "task": StringValue(
+                    description="Task type", default_value="segmentation"
+                ),
+            }
+        )
+        return parameters
+
+    def _normalize(self, tensor: np.ndarray, threshold: float) -> np.ndarray:
+        """Currently supports only min-max normalization."""
+        normalized = ((tensor - threshold) / (self.max - self.min)) + 0.5
```
Wovchena, line 132: This is the only usage of `self.max` and `self.min`. Consider merging them into one parameter since you only need `self.max - self.min`

Wovchena (changes requested):

model_api/python/openvino/model_api/models/anomaly.py
ashwinvaidya17: This is a temporary method. In the next version of anomalib, we are going to have all the post-processing steps as part of the model graph.
The final script will look like this:
image
So I don't want to touch other files to reduce changes. But if you prefer keeping this separate for this version then I can move it

model_api/python/openvino/model_api/models/anomaly.py
```diff
@@ -0,0 +1,152 @@
...
+        super().__init__(inference_adapter, configuration, preload)
+        # attributes for mypy
+        self.max: float
+        self.min: float
+        self.image_threshold: float
+        self.pixel_threshold: float
+        self.task: str
+
+    def preprocess(self, inputs: np.ndarray):
+        inputs = inputs / 255.0  # model expects inputs in range [0, 1]
```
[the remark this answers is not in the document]
ashwinvaidya17: This one is a bit tricky. We do use `mean_values` and `scale_values` but they are set to imagenet normalization statistics. The model expects values scaled to `[0,1]`. The pre-processing then applies imagenet mean and scale values.

model_api/python/openvino/model_api/models/anomaly.py
ashwinvaidya17: The third one is `classification`. That is the `len(predictions.shape=1` case. Currently, the anomaly models generate bounding boxes from the predicted masks. Hence, some parts of the processing are common.

Wovchena (changes requested):

model_api/python/openvino/model_api/models/anomaly.py
```diff
@@ -0,0 +1,149 @@
...
+        """Post-processes the outputs and returns the results.
+
+        Args:
+            outputs (Dict[str, np.ndarray]): Raw model outputs
+            meta (Dict[str, Any]): Meta data containing the original image shape
+
+        Returns:
+            AnomalyResult: Results
+        """
+        anomaly_map: np.ndarray | None = None
```
Wovchena, line 68: That's mostly up to you, it's just a suggestion. C++ version handles this by having empty cv::Mat. Analog to this in python is to have np.ndarray(0), not None.

model_api/python/openvino/model_api/models/anomaly.py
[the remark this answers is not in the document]
Wovchena: C++ test fails. You should temporary add `if` which would just return similar to https://github.com/openvinotoolkit/model_api/pull/93/files#diff-defd17ca74913ca0d56d348f2e3157482ecc581b168194b77368e5daead0f45fR88

Wovchena, line 1: I forgot to ask about it before. Please extend https://github.com/openvinotoolkit/model_api/blob/master/docs/model-configuration.md. You can do it here on in C++ PR

model_api/python/openvino/model_api/models/anomaly.py
```diff
@@ -0,0 +1,152 @@
...
+from .types import ListValue, NumericalValue, StringValue
+from .utils import AnomalyResult
+
+
+class AnomalyDetection(ImageModel):
+    __model__ = "AnomalyDetection"
+
+    def __init__(self, inference_adapter, configuration=None, preload=False):
+        super().__init__(inference_adapter, configuration, preload)
+        # attributes for mypy
```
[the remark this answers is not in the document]
ashwinvaidya17: I think this is an older version of the diff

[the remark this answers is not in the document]
Wovchena: The comment is old, but there's still `            assert anomaly_map is not None  # for mypy`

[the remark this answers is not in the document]
ashwinvaidya17: Ah right. I was looking at the line in this diff. Thanks

Wovchena (approved):

PR #98, opened 2023-07-07
Add Anomaly C++ Wrapper 

ashwinvaidya17:
## Description
- Add Anomaly Wrapper

## Outputs
image


ashwinvaidya17:
@Wovchena I think it is complete. I have the same problem here as the python code as well. The model expects inputs in range [0,1] and the mean and scale values are used for further imagenet pre-processing. For my editor formats everything to 80 characters. Should it be 120 characters? Imo 80 is a bit hard to read.

Wovchena (changes requested):

tests/python/accuracy/public_scope.json
```diff
@@ -304,7 +304,7 @@
             {
                 "image": "coco128/images/train2017/000000000074.jpg",
                 "reference": [
-                    "anomaly_map min:0.49540239572525024 max:1.0;pred_score:1.0;pred_label:Anomalous;pred_mask min:0 max:1;"
+                    "anomaly_map min:0.495402 max:1.0;pred_score:1.0;pred_label:Anomaly;pred_mask min:0 max:1;"
```
Wovchena, line 307: Can you limit the precision. We usually set 3 digits after `.`. The reason is that the exact prediction value is likely to change after new OpenVINO is released

model_api/cpp/models/src/anomaly_model.cpp
```diff
@@ -0,0 +1,230 @@
...
+  cv::Mat pred_mask;
+  std::vector<cv::Rect> pred_boxes;
+  if (predictions.get_shape().size() == 1)
+  {
+    pred_score = predictions.data<float>()[0];
+  }
+  else
+  {
+    anomaly_map = cv::Mat(imageShape[0], imageShape[1], CV_32FC1,
+                          predictions.data<float>());
```
Wovchena, line 88: `anomaly_map` is a `cv::Mat` representation of `predictions`. `predictions` has it's shape, but here you override it with `imageShape`. Please, use `predictions`'s shape. This means you shouldn't have `imageShape` at all

model_api/cpp/models/include/models/anomaly_model.h
```diff
@@ -0,0 +1,63 @@
...
+                                  std::unique_ptr<AnomalyModel> &model);
+
+  static std::string ModelType;
+
+protected:
+  std::vector<int> imageShape = {256, 256};
+  float imageThreshold{0.5f};
+  float pixelThreshold{0.5f};
+  float normalizationScale{1.0f};
+  std::vector<float> mean_values;
```
Wovchena, line 53: Please, move `mean_values` to `ImageModel`, because `ImageModel` is who actually cares about them

model_api/cpp/models/src/anomaly_model.cpp
```diff
@@ -0,0 +1,230 @@
...
+#include "utils/slog.hpp"
+
+std::string AnomalyModel::ModelType = "AnomalyDetection";
+
+/// @brief Initializes the model from the given configuration
+/// @param top_priority  Uses this as the primary source for setting the parameters
+/// @param mid_priority Fallback source for setting the parameters
+void AnomalyModel::init_from_config(const ov::AnyMap &top_priority, const ov::AnyMap &mid_priority)
+{
+  mean_values = get_from_any_maps("mean_values", top_priority, mid_priority, mean_values);
```
Wovchena, line 37: You still use 4 spaces sometimes

model_api/cpp/models/src/anomaly_model.cpp
```diff
@@ -0,0 +1,230 @@
...
+  scale_values = get_from_any_maps("scale_values", top_priority, mid_priority, scale_values);
```
Wovchena, line 38:
```suggestion
```
It's already handled by `ImageModel`

model_api/cpp/models/src/anomaly_model.cpp
```diff
@@ -0,0 +1,230 @@
...
+    pred_boxes = getBoxes(pred_mask);
+  }
+
+  AnomalyResult *result =
+      new AnomalyResult(infResult.frameId, infResult.metaData);
+  result->anomaly_map = anomaly_map;
+  result->pred_score = pred_score;
+  result->pred_label = pred_label;
+  result->pred_mask = pred_mask;
+  result->pred_boxes = pred_boxes;
```
Wovchena, line 121: You should either `std::move()` these values (except for `pred_score`) into `result->` or just use `result->*` directly without construction of intermediate vars throughput this func

model_api/cpp/models/src/anomaly_model.cpp
```diff
@@ -0,0 +1,230 @@
...
+void AnomalyModel::init_from_config(const ov::AnyMap &top_priority, const ov::AnyMap &mid_priority)
+{
+  mean_values = get_from_any_maps("mean_values", top_priority, mid_priority, mean_values);
+  scale_values = get_from_any_maps("scale_values", top_priority, mid_priority, scale_values);
+  imageShape = get_from_any_maps("image_shape", top_priority, mid_priority, imageShape);
+  imageThreshold = get_from_any_maps("image_threshold", top_priority, mid_priority, imageThreshold);
+  pixelThreshold = get_from_any_maps("pixel_threshold", top_priority, mid_priority, pixelThreshold);
+  normalizationScale = get_from_any_maps("normalization_scale", top_priority, mid_priority, normalizationScale);
+  task = get_from_any_maps("task", top_priority, mid_priority, task);
+  labels = get_from_any_maps("labels", top_priority, mid_priority, labels);
```
Wovchena, line 44:
```suggestion
```
It's already handled by `ImageModel`

Wovchena (changes requested):

model_api/cpp/models/src/anomaly_model.cpp
ashwinvaidya17: I have this clangd plugin in my ide. Not sure if it is working correctly.

model_api/cpp/models/src/anomaly_model.cpp
Wovchena: As I stated before, `pred_score` is an exception and shouldn't be `move`ed.

model_api/cpp/models/src/anomaly_model.cpp
Wovchena: There are still lots of them. One example is `  class Model;` in
```cpp
namespace ov
{
  class Model;
} // namespace ov
```

model_api/cpp/models/src/anomaly_model.cpp
Wovchena: No, strings are movable. The criteria is existenc of move constructor, which is (8) in https://en.cppreference.com/w/cpp/string/basic_string/basic_string

ashwinvaidya17: Alright thanks! I think I'll go study C++ :)

Issue #99, opened 2023-07-10
How working `fit_to_window`?


sungchul2:
Hi.
I want to resize an image keeping the aspect ratio, so I used `fit_to_window` because `keep_aspect_ratio=True` in `resize_image_graph` and looks like using `scales` based on the longest axis.
I expected one of two results when resizing from (768, 1024) to (1024, 1024) like the below:
1. (768, 1024) -> (768, 1024)
![image](https://github.com/openvinotoolkit/model_api/assets/97221135/1d657354-6c0f-41c3-aa91-e147b9379e54)

2. (768, 1024) -> (1024, 1024) ((768, 1024) + padding on the bottom (256, 1024) unlike `fit_to_window_letterbox`)
![image](https://github.com/openvinotoolkit/model_api/assets/97221135/37dd8737-ee8e-4d33-97aa-c0f2ae21a573)

But, I got the result (1024, 1024) just resizing an image without any padding.
The aspect ratio of the result was not maintained.
![image](https://github.com/openvinotoolkit/model_api/assets/97221135/e24615cf-cd39-4770-9e6e-9547018175de)

I don't know the difference between `fit_to_window` and `standard`, except for using `scales` instead of `target_size`.
I think changing `shape_calculation_mode` to `scales` or setting `pads_begin` or `pads_end` is needed.
Is it correct?

sungchul2:
```python
import torch
import subprocess
import numpy as np
from openvino.model_api.adapters import OpenvinoAdapter, create_core

class DummyModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.conv = torch.nn.Conv2d(3, 3, (1, 1))

    def forward(self, inputs):
        dummy_output = self.conv(inputs)
        return inputs, dummy_output

model = DummyModel()
torch.onnx.export(model, (torch.zeros(1, 3, 1024, 1024, dtype=torch.float32)), open("test_preprocessor.onnx", "wb"), input_names=["inputs"], output_names=["inputs", "dummy_output"])
subprocess.run(["mo", "--input_model", "test_preprocessor.onnx", "--output_dir", ".", "--model_name", "test_preprocessor"])

model_adapter = OpenvinoAdapter(
    core=create_core(),
    model="test_preprocessor.xml",
    weights_path="test_preprocessor.bin",
    device="CPU",
    max_num_requests=1,
    plugin_config={"PERFORMANCE_HINT": "THROUGHPUT"},
)

[...]

PR #100, opened 2023-07-10
Fix `fit_to_window` Python embed_preprocessing()

Wovchena:
Close https://github.com/openvinotoolkit/model_api/issues/99

The implementation was incorrect. It performed scales based interpolation instead if sizes. Padding was missing. Now it is aligned with C++.


sungchul2:
Thanks for quick fix.
Although I didn't check, some functions use metadata related to `fit_to_window` and used into `postprocess` might be affected, such as
https://github.com/openvinotoolkit/model_api/blob/c8280503ba5642d32b77b67f661049c1e17ee74d/model_api/python/openvino/model_api/models/detection_model.py#L136-L146
What do you think?

Wovchena:
Updated

sungchul2 (approved):

PR #106, opened 2023-07-19
Add handing of onnx metadata

sovrasov:
# What does this PR do?

This PR slightly modifies python OV adapter to read an analog of rt_info from onnx.
Technical details:
- rt_info is stored in onnx_model.metadata_props
- key of the metadata property is a string containing space-separated xml tags from the original rt_info path, for instance, `model_info confidence_threshold`, `model_info hierarchical_config`. All values of attrs are stored as strings, as in rt_info

Fixes # (issue)

## Before submitting


sovrasov:
Implementation is finished, now waiting for the model to be uploaded to s3.

model_api/python/openvino/model_api/adapters/openvino_adapter.py
```diff
@@ -18,13 +18,16 @@
 from pathlib import Path
 from typing import Dict, Set, Tuple
 
+import onnx
```
Wovchena, line 21: Please, move the import under `if Path(self.model_path).suffix == ".onnx" and not weights_path:` to make it optional

Wovchena (changes requested):

model_api/python/openvino/model_api/adapters/openvino_adapter.py
sovrasov: Done

Wovchena (approved):

sovrasov:
@Wovchena tests are passing now

PR #112, opened 2023-07-24
Add probs output to satisfy Geti's research framework reqs

sovrasov:
# What does this PR do?

In C++ implementation raw scores are not required at the moment

Fixes # (issue)

## Before submitting

### DIFF
--- a/model_api/python/openvino/model_api/models/classification.py
+++ b/model_api/python/openvino/model_api/models/classification.py
@@ -37,12 +37,14 @@ class ClassificationModel(ImageModel):
         if 1 == len(self.outputs):
             self._verify_signle_output()
 
+        self.raw_scores_name = _raw_scores_name
         if self.hierarchical:
             self.embedded_processing = True
             self.out_layer_names = _get_non_xai_names(self.outputs.keys())
             _append_xai_names(self.outputs.keys(), self.out_layer_names)
             if not self.hierarchical_config:
                 self.raise_error("Hierarchical classification config is empty.")
+            self.raw_scores_name = self.out_layer_names[0]
             self.hierarchical_info = json.loads(self.hierarchical_config)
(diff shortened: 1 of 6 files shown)


model_api/python/openvino/model_api/models/classification.py
```diff
@@ -154,12 +154,39 @@ def postprocess(self, outputs, meta):
         else:
             result = self.get_multiclass_predictions(outputs)
 
+        raw_scores = np.ndarray(0)
+        if self.output_raw_scores:
+            raw_scores = self.get_all_probs(outputs[self.out_layer_names[-1]])
```
Wovchena, line 159:
```suggestion
            raw_scores = self.get_all_probs(outputs[_raw_scores_name])
```
When I was fixing ModelAPI for direct adapter usage, I noticed that the order of output names may change (which is valid, because the opposite is never promised).

model_api/python/openvino/model_api/models/classification.py
Wovchena, line 1: > In C++ implementation raw scores are not required at the moment

I'm OK to merge Python only. At the same time I'd appreciate if you updated C++ at spare time as well. This will keep implementations aligned and also this should be simpler now, because you still remember all the context of this change for Python.

Wovchena (changes requested):

model_api/python/openvino/model_api/models/classification.py
sovrasov: I'll do that in this PR

model_api/python/openvino/model_api/models/classification.py
sovrasov: Updated

model_api/python/openvino/model_api/models/classification.py
sovrasov: Done

PR #118, opened 2023-08-04
Fix cls transformers inference

sovrasov:
# What does this PR do?

Softmax node now is searched in the output nodes only

Fixes # (issue)

## Before submitting

### DIFF
--- a/model_api/python/openvino/model_api/models/classification.py
+++ b/model_api/python/openvino/model_api/models/classification.py
@@ -60,10 +60,9 @@ class ClassificationModel(ImageModel):
                 self.load()
             return
 
-        if not self.embedded_processing:
-            addOrFindSoftmaxAndTopkOutputs(
-                self.inference_adapter, self.topk, self.output_raw_scores
-            )
+        addOrFindSoftmaxAndTopkOutputs(
+            self.inference_adapter, self.topk, self.output_raw_scores
+        )
         self.embedded_processing = True
 
         self.out_layer_names = ["indices", "scores"]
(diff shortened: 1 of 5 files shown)


sovrasov:
@Wovchena are you ok with this approach? If yes, I'll try to implement it in cpp as well

model_api/python/openvino/model_api/models/classification.py
```diff
@@ -243,9 +243,13 @@ def get_multiclass_predictions(self, outputs):
...
     nodes = inference_adapter.model.get_ops()
     softmaxNode = None
+    all_output_names = []
+    for output in inference_adapter.model.outputs:
+        all_output_names.extend(output.names)
     for op in nodes:
         if "Softmax" == op.get_type_name():
-            softmaxNode = op
+            if op.get_friendly_name() in all_output_names:
+                softmaxNode = op
```
Wovchena, line 252: `ov::Model` has `const ov::ResultVector& get_results()` where `using ResultVector = std::vector<std::shared_ptr<ov::op::v0::Result>>;` and `Result` is a subclass of `Op` which is a subclass of `Node`.

Can you try replacing `nodes = inference_adapter.model.get_ops()` with `inference_adapter.model.get_results()`

Wovchena (changes requested):

model_api/python/openvino/model_api/models/classification.py
Wovchena: Well, I guess we can consider there is no other way. So I'm ok with this approach. Just make `all_output_names` to be a set, not list

sovrasov: I implemented another version of workaround, @Wovchena please have a look

tests/python/accuracy/public_scope.json
```diff
@@ -292,7 +292,7 @@
             {
                 "image": "coco128/images/train2017/000000000074.jpg",
                 "reference": [
-                    "anomaly_map min:0.5944074988365173 max:1.0;pred_score:1.0;pred_label:Anomalous;pred_mask min:1 max:1;"
+                    "anomaly_map min:0.5944 max:1.0000;pred_score:1.0;pred_label:Anomalous;pred_mask min:1.0000 max:1.0000;"
```
Wovchena, line 295: I expected this to happen here: https://github.com/openvinotoolkit/model_api/pull/98#discussion_r1277247488. But the PR seems to be stuck. Can you just change the precision to 3 digits, not 4, to stay consistent with the rest?

sovrasov: Done

Wovchena (approved):

PR #126, opened 2023-08-16
Copy https://github.com/openvinotoolkit/openvino/blob/cabb40638a386d32e704d1001315ded76598d5ff/src/bindings/python/src/openvino/__init__.py

Wovchena:


.github/workflows/test_precommit.yml
```diff
@@ -45,11 +45,11 @@ jobs:
     - name: Check style with black
       run: |
         source venv/bin/activate
-        black --check .
+        black --exclude model_api/python/openvino/__init__.py --check .
```
sovrasov, line 48: Can we add skip directives to the file? Passing ignore list to command line is not very convenient and scalable

Wovchena: black reads `pyproject.toml`. But having `pyproject.toml` would suggest that we support building a package from it

sovrasov (approved):

AlexKoff88:
@ilya-lavrenov, fyi. It seems like the namespace remains the same.

@Wovchena, when are you planning to change the namespace and remove `openvino.` from the beginning?

Wovchena:
Everyone I discussed this topic with prefers to keep `openvino.`, so I'm going to keep it.

AlexKoff88 (approved):

PR #127, opened 2023-08-22
Add ONNX runtime inference adapter

sovrasov:
# What does this PR do?

This PR adds ONNX runtime -based inference adapter. Currently tested only with multilabel classification model

```python
onnx_adapter = ONNXRuntimeAdapter("model.onnx")
model_onnx = ClassificationModel.create_model(onnx_adapter)
prediction = model_onnx(image)
```

Fixes # (issue)

## Before submitting


sovrasov:
This PR misses a lot of details, but it showcases the approach: use python preprocessing and onnx metadata.
@Wovchena @AlexKoff88 please have a look if non-OV adapter is acceptable from the overall conception.
Unfortunately, some OV stuff leaked to implementation of particular models (like add softmax for classification), and this should be addressed in a modelwise manner.

Wovchena:
What's the reason to have onnx backed? You've already enabled inference by https://github.com/openvinotoolkit/model_api/pull/106

sovrasov:
> What's the reason to have onnx backed? You've already enabled inference by #106

Some of the Geti customers prefer to use ONNX runtime rather than OV for inference on their side

Wovchena (approved):

Wovchena:
Can you add a note that people shouldn't expect it to work in every case, so I could ignore bugs related to this.

sovrasov:
> Can you add a note that people shouldn't expect it to work in every case, so I could ignore bugs related to this.

Sure, I'll add this.

AlexKoff88:
For my reference, who needs ORT inference and for which workloads?

model_api/python/openvino/model_api/adapters/utils.py
```diff
@@ -378,3 +378,35 @@ def resize_image_letterbox(size, interpolation, pad_value):
             pad_value=pad_value,
         )
     )
+
+
+def load_parameters_from_onnx(onnx_model):
+    parameters = {}
+
+    def insert_hiararchical(keys, val, root_dict):
```
AlexKoff88, line 386:
```suggestion
    def insert_hierarchical(keys, val, root_dict):
```

model_api/python/openvino/model_api/adapters/utils.py
```diff
@@ -378,3 +378,35 @@ def resize_image_letterbox(size, interpolation, pad_value):
...
+        if len(keys) == 1:
+            root_dict[keys[0]] = val
+            return
+        if keys[0] not in root_dict:
+            root_dict[keys[0]] = {}
+        insert_hiararchical(keys[1:], val, root_dict[keys[0]])
```
AlexKoff88, line 392:
```suggestion
        insert_hierarchical(keys[1:], val, root_dict[keys[0]])
```

AlexKoff88:
Need a couple of tests that verify that it works as expected to the supported cases and throughs an exception for the unsupported model types.

model_api/python/openvino/model_api/adapters/utils.py
sovrasov: Fixed

sovrasov:
> Need a couple of tests that verify that it works as expected to the supported cases and throughs an exception for the unsupported model types.

The tests were added, similar to other adapters. Unfortunately, the inference results are slightly different than ones obtained with OV. Since we can't change anything in the inference process itself, the next step would be to compare OV vs python preprocessing. We already faced that problem when upgraded from old OMZ model wrappers in OTX. I'm planning to add testcases that measure the exact difference. 

@Wovchena it seems like we can't add implicit assertions to prohibit usage of ORT adapter with untested wrappers, because those model wrappers should know about ORT or any other adapters (in case of a whitelist). Maybe I overlooked an obvious solution though.

Wovchena:
Model's `__init__()` takes `inference_adapter`. You can try verifying the type for each subclass.

sovrasov:
> Model's `__init__()` takes `inference_adapter`. You can try verifying the type for each subclass.

Yep, that's a possible option, but we have to import all wrappers in model.py. Is that fine?

Wovchena:
You can perform the check in each subclass

model_api/python/openvino/model_api/models/model.py
```diff
@@ -76,6 +76,19 @@ def __init__(self, inference_adapter, configuration=dict(), preload=False):
...
         self.logger = log.getLogger()
         self.inference_adapter = inference_adapter
+        if not isinstance(
+            self.inference_adapter, (OpenvinoAdapter, OVMSAdapter)
+        ) and self.__model__ not in [
+            "Classification",
+            "MaskRCNN",
+            "SSD",
+            "Segmentation",
+        ]:
```
Wovchena, line 86:
```suggestion
        ) and self.__model__ not in {
            "Classification",
            "MaskRCNN",
            "SSD",
            "Segmentation",
        }:
```

model_api/python/openvino/model_api/models/model.py
```diff
@@ -76,6 +76,19 @@ def __init__(self, inference_adapter, configuration=dict(), preload=False):
...
+            raise WrapperError(
+                f"{self.__model__}",
```
Wovchena, line 88:
```suggestion
            self.raise_error(
```
To stay consistent

Wovchena (changes requested):

model_api/python/openvino/model_api/models/model.py
sovrasov: Updated

sovrasov:
@AlexKoff88 @Wovchena the tests & fixes are ready. Please have a look one again if this can be merged

model_api/python/openvino/model_api/adapters/onnx_adapter.md
```diff
@@ -0,0 +1,56 @@
...
+### ONNX metadata
+
+ModelAPI uses IR RTInfo to store metadata (wrapper-specific parameters, preprocessing parameters, labels list, etc.).
+For details see the implementation of `Model._load_config()` method.
+To embed that metadata into ONNX file, one can use ONNX properties: `metadata_props.add()` and use
+ModelAPI-specific parameters as metadata keys with exactly the same names as in RTInfo, but split by spaces: `"model_info model_type"` and so on.
+
+## Limitations
+
+- Although `ONNXRuntimeAdapter` doesn't use OpenVINO directly, OV should be installed, because Model API depends on it at the low level.
```
AlexKoff88, line 21: OpenVINO is in the mandatory dependencies. I don't think you should mention it here.

AlexKoff88:
I am ok with the changes

AlexKoff88 (approved):

model_api/python/openvino/model_api/adapters/onnx_adapter.md
```diff
@@ -0,0 +1,56 @@
...
+- Model reshape is not supported, and input shape should be define in the model (excluding batch dimension) to perform shape inference and
```
Wovchena, line 22:
```suggestion
- Model reshape is not supported, and input shape should be defined in the model (excluding batch dimension) to perform shape inference and
```

model_api/python/openvino/model_api/adapters/onnx_adapter.py
```diff
@@ -0,0 +1,207 @@
...
+class ONNXRuntimeAdapter(InferenceAdapter):
+    """
+    This inference adapter allows running ONNX models via ONNXRuntime.
+    The adapter has limited functionality: it supports only image models
+    generated by OpenVINO training extensions (OTX). Each onnx file
+    generated by OTX contains ModelAPI-style metadata, which is used for
+    configuring a particular model acting on top of it.
+    Also, this adapter doesn't provide asynchronous inference functionality and model reshaping.
+    Visit https://github.com/openvinotoolkit/training_extensions/ for the
+    comprehensive list of models available.
```
Wovchena, line 46: Here you refer to OTX, but in readme you list
> `SSD`, `MaskRCNNModel`, `SegmentationModel`, and `ClassificationModel` wrappers.

Please, align the messages

model_api/python/openvino/model_api/models/classification.py
```diff
@@ -60,14 +60,21 @@ def __init__(self, inference_adapter, configuration=dict(), preload=False):
...
+                self.inference_adapter, self.topk, self.output_raw_scores
+            )
+            self.embedded_topk = True
+            self.out_layer_names = ["indices", "scores"]
+            if self.output_raw_scores:
+                self.out_layer_names.append(self.raw_scores_name)
+        except (RuntimeError, AttributeError):
+            self.embedded_topk = False
+            self.out_layer_names = _get_non_xai_names(self.outputs.keys())
+            self.raw_scores_name = self.out_layer_names[0]
```
Wovchena, line 74: Just a note, if you are ever going to modify `classification.py` further, it may be simpler to remove ngraph based postprocessing from C++ and Python

model_api/python/openvino/model_api/adapters/onnx_adapter.md
Wovchena, line 21: Can you address
> I would note instead that there is only Python version of ORT adapter.

Regarding
> OpenVINO is in the mandatory dependencies. I don't think you should mention it here.

I think it won't harm to explain why OV is always required, and given @AlexKoff88 approved the current state, I think we can ignore that :)

Wovchena (changes requested):

model_api/python/openvino/model_api/models/classification.py
sovrasov: Ok, let's see if this fallback logic requires any modification in the future.

model_api/python/openvino/model_api/adapters/onnx_adapter.md
sovrasov: Done

model_api/python/openvino/model_api/adapters/onnx_adapter.md
sovrasov: Fixed

model_api/python/openvino/model_api/adapters/onnx_adapter.py
sovrasov: Updated: particular adapters are mentioned now, like in reamde

model_api/python/openvino/model_api/adapters/onnx_adapter.md
sovrasov: > I would note instead that there is only Python version of ORT adapter.

Done.

Regarding the deps notice, it was targeted to onnx fans among Geti users: from their side it's not that clear why do we need to install OV if the inference is conducted via ONNXRT.

PR #134, opened 2023-09-20
Explicit rotated_rect, simplify example

Wovchena:


examples/cpp/synchronous_api/main.cpp
```diff
@@ -32,41 +32,32 @@
...
-        }
-
-    } catch (const std::exception& error) {
-        std::cerr << error.what() << std::endl;
-        return 1;
-    } catch (...) {
-        std::cerr << "Unknown/internal exception happened." << std::endl;
-        return 1;
+            << " | " << std::setw(4) << int(obj.x) << " | " << std::setw(4) << int(obj.y) << " | "
+            << std::setw(4) << int(obj.x + obj.width) << " | " << std::setw(4) << int(obj.y + obj.height) << "\n";
```
sovrasov, line 55: Do we need `return 0;` here ?

Wovchena: It turns out main is so much special that it's not required to return 0. https://en.cppreference.com/w/cpp/language/main_function

sovrasov (approved):

PR #139, opened 2023-10-13
Remove classification model patching

sovrasov:
# What does this PR do?
This PR removes insertion of Softmax/topk postprocessing to the OV model.
Instead, only presence of softmax is detected to avoid double activation.

Fixes # (issue)

## Before submitting


Wovchena:
Are you going to modify the wrapper in the upcoming patches further? I mean it would be strange to modify it here just out of blue

sovrasov:
I was going to update it to allow patching of output_raw_scores parameter after a model is created in python, but it seems like I don't need it anymore.

BTW the method I use here doesn't work in cpp, so let's close this PR for now.

PR #140, opened 2023-10-20
Fix width and height extraction

Wovchena:
Ticket 123278

Currently there's no OMZ model adapted for model_api which would trigger such case, thus I'm not adding a test.


ljcornel (approved):
Looks good to me!

PR #143, opened 2023-11-30
SegmentationModel - get proper output in postprocess()

mgumowsk:
# What does this PR do?

Using SegmentationModel with OpenVINO Model Server adapter caused ov::merge exception.
The reason is the order of the outputs.
This PR makes sure, that `feature_vector` is not used in postprocess() function.


sovrasov (approved):

mgumowsk:
> @mgumowsk we can merge this if it's urgent, since #144 fails on the same testcases. I believe that happened because of a new version of ultralytics package and OV models converted with a new OV version

@sovrasov if you can then yes, please merge, without it segmentation model doesn't work in Geti.

PR #148, opened 2023-12-19
Add support of models converted by OVC

sovrasov:
# What does this PR do?

When loading models one can use indices instead of names

Fixes # (issue)

## Before submitting


sovrasov:
Waiting for the files to appear on s3...

PR #154, opened 2024-01-29
Apply newer black

Wovchena:


ashwinvaidya17 (approved):
Thanks

ashwinvaidya17:
I can see that the accuracy test is failing but I guess it is due to the new models.

PR #159, opened 2024-02-22
Fix saliency maps order for h-cls in cpp

sovrasov:
# What does this PR do?

The order could have been twisted, because logits in hierarchical classification are not placed in the same order as "labels" in model info.

Fixes # (issue)

## Before submitting


model_api/cpp/models/src/classification_model.cpp
```diff
@@ -358,6 +359,27 @@ std::unique_ptr<ResultBase> ClassificationModel::get_hierarchical_predictions(In
     return retVal;
 }
 
+ov::Tensor ClassificationModel::reorder_saliency_maps(const ov::Tensor& source_maps) {
+    if (!hierarchical || source_maps.get_shape().size() == 1) {
```
GalyaZalesskaya, line 363: In which case source_maps.get_shape().size()==1 can be  true?

model_api/cpp/models/src/classification_model.cpp
```diff
@@ -358,6 +359,27 @@ std::unique_ptr<ResultBase> ClassificationModel::get_hierarchical_predictions(In
...
+
+    auto reordered_maps = ov::Tensor(source_maps.get_element_type(), source_maps.get_shape());
+    const std::uint8_t* source_maps_ptr = static_cast<std::uint8_t*>(source_maps.data());
+    std::uint8_t* reordered_maps_ptr = static_cast<std::uint8_t*>(reordered_maps.data());
+
+    size_t shape_offset = (source_maps.get_shape().size() == 4) ? 1 : 0;
+    size_t map_byte_size = source_maps.get_element_type().size() *
+        source_maps.get_shape()[shape_offset + 1] * source_maps.get_shape()[shape_offset + 2];
+
+    for (size_t i = 0; i < source_maps.get_shape()[shape_offset]; ++i) {
```
GalyaZalesskaya, line 375: Does it work if there are several maps returned, so source_maps have batch size 2? For instance (2,3,7,7) for 2 images and 3 classes

model_api/cpp/models/src/classification_model.cpp
sovrasov: If saliency map is an empty tensor, that's a sanity check

GalyaZalesskaya (approved):

sovrasov:
cpp part is finem so we can merge this

PR #163, opened 2024-02-27
Move to newer OMZ

Wovchena:
After this is merged It will be required to release model_api 0.2.0 only after openvino 2024.0 because model_api is going to import OMZ using a new namespace

### DIFF
--- a/model_api/python/requirements.txt
+++ b/model_api/python/requirements.txt
@@ -1,5 +1,7 @@
 numpy>=1.16.6
 opencv-python
 scipy>=1.5.4
-openvino>=2023.0.0
+# TODO: set 2024.0 after the release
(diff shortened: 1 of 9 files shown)


model_api/python/requirements.txt
```diff
@@ -1,5 +1,7 @@
 numpy>=1.16.6
 opencv-python
 scipy>=1.5.4
-openvino>=2023.0.0
+# TODO: set 2024.0 after the release
+openvino>=2023.3
 openvino-dev>=2023.0.0
```
sovrasov, line 6: 2023.3.0?

Wovchena: It works both ways

PR #174, opened 2024-05-21
Make tile with full img optional

eugene123tw:
# What does this PR do?

Provide an option to include the full image on the tile.

## Before submitting

### DIFF
--- a/model_api/python/model_api/tilers/detection.py
+++ b/model_api/python/model_api/tilers/detection.py
@@ -123,11 +123,8 @@ class DetectionTiler(Tiler):
                 iou_threshold=self.iou_threshold,
             )
 
-        start_idx = 1 if self.tile_with_full_img else 0
         merged_vector = (
-            np.mean(feature_vectors[start_idx:], axis=0)
-            if feature_vectors
-            else np.ndarray(0)
+            np.mean(feature_vectors, axis=0) if feature_vectors else np.ndarray(0)
         )
         saliency_map = (
             self._merge_saliency_maps(saliency_maps, shape, tiles_coords)
--- a/model_api/python/model_api/tilers/detection.py
+++ b/model_api/python/model_api/tilers/detection.py
@@ -125,7 +125,9 @@ class DetectionTiler(Tiler):
 
         start_idx = 1 if self.tile_with_full_img else 0
         merged_vector = (
-            np.mean(feature_vectors[start_idx:], axis=0) if feature_vectors else np.ndarray(0)
+            np.mean(feature_vectors[start_idx:], axis=0)
+            if feature_vectors
+            else np.ndarray(0)
         )
         saliency_map = (
             self._merge_saliency_maps(saliency_maps, shape, tiles_coords)
@@ -185,7 +187,9 @@ class DetectionTiler(Tiler):
         image_map_w = int(image_w * ratio[1])
         merged_map = np.zeros((num_classes, image_map_h, image_map_w))
 
-        saliency_maps, start_idx = (saliency_maps[1:], 1) if self.tile_with_full_img else (saliency_maps, 0)
+        saliency_maps, start_idx = (
+            (saliency_maps[1:], 1) if self.tile_with_full_img else (saliency_maps, 0)
+        )
         for i, saliency_map in enumerate(saliency_maps, start_idx):
             for class_idx in range(num_classes):
                 if len(saliency_map.shape) == 4:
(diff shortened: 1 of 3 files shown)


model_api/python/model_api/tilers/detection.py
```diff
@@ -123,8 +123,9 @@ def _merge_results(self, results, shape):
                 iou_threshold=self.iou_threshold,
             )
 
+        start_idx = 1 if self.tile_with_full_img else 0
```
sovrasov, line 126: Why do we need these changes? It seems like removing the first tile in model_api/python/model_api/tilers/tiler.py is enough

model_api/python/model_api/tilers/detection.py
```diff
@@ -123,8 +123,9 @@ def _merge_results(self, results, shape):
...
         merged_vector = (
-            np.mean(feature_vectors, axis=0) if feature_vectors else np.ndarray(0)
+            np.mean(feature_vectors[start_idx:], axis=0) if feature_vectors else np.ndarray(0)
```
sovrasov, line 128: Why do we need to change the logic for feature_vector?

model_api/python/model_api/tilers/detection.py
```diff
@@ -184,7 +185,8 @@ def _merge_saliency_maps(self, saliency_maps, shape, tiles_coords):
         image_map_w = int(image_w * ratio[1])
         merged_map = np.zeros((num_classes, image_map_h, image_map_w))
 
-        for i, saliency_map in enumerate(saliency_maps[1:], 1):
+        saliency_maps, start_idx = (saliency_maps[1:], 1) if self.tile_with_full_img else (saliency_maps, 0)
+        for i, saliency_map in enumerate(saliency_maps, start_idx):
```
GalyaZalesskaya, line 189:
```suggestion
        start_idx = 1 if self.tile_with_full_img else 0
        for i, saliency_map in enumerate(saliency_maps[start_idx:]):
```
We can simplify that since `i` is just an index to iterate through list of saliency maps without any additional meaning

model_api/python/model_api/tilers/instance_segmentation.py
```diff
@@ -183,7 +184,8 @@ def _merge_saliency_maps(self, saliency_maps, shape, tiles_coords):
 
         merged_map = [np.zeros((image_h, image_w)) for _ in range(num_classes)]
 
-        for i, saliency_map in enumerate(saliency_maps[1:], 1):
+        saliency_maps, start_idx = (saliency_maps[1:], 1) if self.tile_with_full_img else (saliency_maps, 0)
+        for i, saliency_map in enumerate(saliency_maps, start_idx):
```
GalyaZalesskaya, line 188: Same

model_api/python/model_api/tilers/detection.py
eugene123tw: I have aligned this with the PyTorch implementation: https://github.com/openvinotoolkit/training_extensions/pull/3530/files#diff-3e70350ee460c59caf89aff057b556900998828cf13f7a1ebe802e89068fd6e1R210.

I'm unsure which approach provides the best representation. @GalyaZalesskaya, could you share your opinion?

GalyaZalesskaya: Yes, basically for feature vector it doesn't matter whether we use first tile with the whole image or not. I thought we can align approach for saliency maps (where we use tile with image in a special way) and skip the tile with image for feature vector.

But now it seems to me that for feature vector it's better to leave the initial implementation in PyTorch and OpenVINO.
@eugene123tw Can I ask you to revert changes?

eugene123tw: reverted

GalyaZalesskaya (approved):

model_api/python/model_api/tilers/tiler.py
```diff
@@ -93,6 +93,10 @@ def parameters(cls):
                     max=1.0,
                     description="Overlap of tiles",
                 ),
+                "tile_with_full_img": BooleanValue(
+                    default_value=False,
```
sovrasov, line 97: This flag should be true by default to preserve backward compatibility.

sovrasov: Also, this'd fix the python accuracy test.

eugene123tw: Thanks for the comment, and updated.

sovrasov (approved):

PR #177, opened 2024-05-24
Async inference for cpp

sovrasov:
# What does this PR do?

Asynchronous inference for cpp implementation is added:
- Async queue
- Batch mode emulation
- Async tiling via batch mode
- Example

Batch inference emulation with 100 inputs:
image


model_api/cpp/utils/include/utils/async_infer_queue.hpp
```diff
@@ -0,0 +1,48 @@
...
+    void wait_all();
+    void set_default_callbacks();
+    void set_custom_callbacks(std::function<void(ov::InferRequest, std::shared_ptr<ov::AnyMap> callback_args)> f_callback);
+    size_t size() const;
+    void start_async(const ov::Tensor& input, std::shared_ptr<ov::AnyMap> userdata = nullptr);
+    void start_async(const std::map<std::string, ov::Tensor>& input, std::shared_ptr<ov::AnyMap> userdata = nullptr);
+    ov::InferRequest operator[](size_t i);
+
+    // AsyncInferQueue is the owner of all requests. When AsyncInferQueue is destroyed,
+    // all of requests are destroyed as well.
```
Wovchena, line 41: You can't really control that. `ov::InferRequest` is just a shared_ptr.  One can get an `ov::InferRequest` with `operator[]()` and copy it. After the `m_requests` is destroyed, the internals of `ov::InferRequest` will continue living. I think you should remove the comment or leave an explanation similar to that.

model_api/cpp/utils/src/async_infer_queue.cpp
```diff
@@ -0,0 +1,167 @@
...
+        m_requests.emplace_back(model.create_infer_request());
+        m_user_ids.push_back(nullptr);
+        m_idle_handles.push(handle);
+    }
+
+    set_default_callbacks();
+}
+
+AsyncInferQueue::~AsyncInferQueue() {
+    m_requests.clear();
```
Wovchena, line 40: An empty destructor would destroy `m_requests` anyway. I think you can keep `~AsyncInferQueue() = default;`

model_api/cpp/utils/src/async_infer_queue.cpp
```diff
@@ -0,0 +1,167 @@
...
+            // Notify locks in getIdleRequestId()
+            m_cv.notify_one();
+
+            try {
+                if (exception_ptr) {
+                    std::rethrow_exception(exception_ptr);
+                }
+            } catch (const std::exception& e) {
+                OPENVINO_THROW(e.what());
+            }
```
Wovchena, line 95: The idea of providing `exception_ptr` is to forward the error to the main thread instead of silencing it in a callback thread. You should push it to `m_errors` instead of rethrowing. Same for `set_custom_callbacks()`

model_api/cpp/adapters/src/openvino_adapter.cpp
```diff
@@ -34,23 +44,48 @@ void OpenVINOInferenceAdapter::loadModel(const std::shared_ptr<const ov::Model>&
 }
 
 InferenceOutput OpenVINOInferenceAdapter::infer(const InferenceInput& input) {
+    auto request = (*asyncQueue)[asyncQueue->get_idle_request_id()];
```
Wovchena, line 47:
```suggestion
    auto request = asyncQueue,get()[asyncQueue->get_idle_request_id()];
```
It was hard to read

model_api/cpp/models/src/image_model.cpp
```diff
@@ -87,6 +88,24 @@ ImageModel::ImageModel(std::shared_ptr<InferenceAdapter>& adapter, const ov::Any
...
+    return ModelBase::infer(static_cast<const InputData&>(inputData));;
+}
+
+std::vector<std::unique_ptr<ResultBase>> ImageModel::inferBatchImage(const std::vector<ImageInputData>& inputImgs) {
+    std::vector<std::reference_wrapper<const InputData>> inputData;
+    inputData.reserve(inputImgs.size());
+    for (const auto& img : inputImgs) {
+        inputData.push_back(static_cast<const InputData&>(img));
+    }
+    ModelBase::inferBatch(std::vector<InputData>());
```
Wovchena, line 101: Why do you infer empty data?

Wovchena (changes requested):

model_api/cpp/models/src/image_model.cpp
sovrasov: That's a last-minute refactoring artifact, will delete

model_api/cpp/utils/include/utils/async_infer_queue.hpp
sovrasov: I see, the comments came from the original implementation. I'll update it

model_api/cpp/utils/src/async_infer_queue.cpp
sovrasov: Deleted

model_api/cpp/adapters/src/openvino_adapter.cpp
sovrasov: Changed to `asyncQueue->operator[]()`, because after `.get()` dereferencing is required as well

Wovchena: I thought I deleted that comment :)

model_api/cpp/utils/src/async_infer_queue.cpp
sovrasov: Done

model_api/cpp/adapters/src/openvino_adapter.cpp
```diff
@@ -17,15 +17,25 @@
...
+                                                            const std::string& device, const ov::AnyMap& compilationConfig, size_t max_num_requests) {
     slog::info << "Loading model to the plugin" << slog::endl;
-
-    compiledModel = core.compile_model(model, device, compilationConfig);
-    inferRequest = compiledModel.create_infer_request();
-
+    ov::AnyMap customCompilationConfig(compilationConfig);
+    if (max_num_requests != 1) {
+        if (customCompilationConfig.find("PERFORMANCE_HINT") == customCompilationConfig.end()) {
+            customCompilationConfig["PERFORMANCE_HINT"] = ov::hint::PerformanceMode::THROUGHPUT;
```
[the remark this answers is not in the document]
sovrasov: > Maybe it will even switch latency to throughput automatically.

I haven't found that in the docs, so I experimented a bit.
The optimal number of requests is not known in advance, so basically we have several options:
1) user sets the number explicitly, throughput mode is enabled explicitly if we have more that 1 request
2) user relies on `model.get_property(ov::optimal_number_of_infer_requests)`, and creates that many requests manually:
if throughput mode is not set, then no speedup
3) ov::hint::num_requests is passed, throughput mode is not set:
`model.get_property(ov::optimal_number_of_infer_requests)` is 1. My assumption this is because throughput mode is not set under the hood.
4) Like in docs, `ov::hint::performance_mode(ov::hint::PerformanceMode::THROUGHPUT),
    ov::hint::num_requests(4)` are set. 
`model.get_property(ov::optimal_number_of_infer_requests)` is 4, exactly as suggested in the hint

Among these option I'd prefer 1 and 2 which are already implemented.
Besides, to support `ov::hint`, we need to get rid of AnyMap compilation configuration.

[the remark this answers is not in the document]
Wovchena: What's wrong with (4)? I believe that should be preferred. My guess is that `model.get_property(ov::optimal_number_of_infer_requests)` is 4 because `num_requests(4)` sets throughput to use 4 streams. If you don't provide `num_requests`, the number of streams is equal to the number of threads for `CPU`. Using 4 ireqs only leaves some of the streams unused.

model_api/cpp/utils/src/async_infer_queue.cpp
Wovchena: Can you move to the top of callback. It's pointless to process an output if inference failed

PR #178, opened 2024-05-29
Add action classification class

eunwoosh:
# What does this PR do?
Add ActionClassificationModel which is a wrapper class for action classification model.

## Before submitting
- [x] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests?

### DIFF
--- a/model_api/python/docs/ActionClassification.md
+++ b/model_api/python/docs/ActionClassification.md
@@ -1,0 +1,79 @@
+# ActionClassification Wrapper
+
+## Use Case and High-Level Description
+
+The `ActionClassificationModel` is a wrapper class designed for action classification models.
+This class provides support for data preprocessing and postprocessing like other model wrapper classes.
+Note that it isn't a subclass of `ImageModel`. It gets video as input so it is different than ImageModel.
+
+## How to use
+
+Utilizing the ActionClassificationModel is similar to other model wrappers, with the primary difference being the preparation of video clip inputs instead of single images.
+
+Below is an example demonstrating how to initialize the model with OpenVINO™ IR files and classify actions in a video clip.
+
+
+```python
+import cv2
+import numpy as np
+# import model wrapper class
+from model_api.models import ActionClassificationModel
+# import inference adapter and helper for runtime setup
+from model_api.adapters import OpenvinoAdapter, create_core
+
+
+# load video and make a clip as input
+cap = cv2.VideoCapture("sample.mp4")
+input_data = np.stack([cap.read()[1] for _ in range(8)])
+
+
+# define the path to action classification model in IR format
+model_path = "action_classification.xml"
+
+# create adapter for OpenVINO™ runtime, pass the model path
+inference_adapter = OpenvinoAdapter(create_core(), model_path, device="CPU")
+
+# instantiate the ActionClassificationModel wrapper
+# setting preload=True loads the model onto the CPU within the adapter0
+action_cls_model = ActionClassificationModel(inference_adapter, preload=True)
+
+# perform preprocessing, inference, and postprocessing
+results = action_cls_model(input_data)
+```
+
+As illustrated, initializing the model and performing inference can be achieved with minimal code.
+The wrapper class takes care of input processing, layout adjustments, and output processing automatically.
+
+
+## Arguments
+
+- `labels`(`list[str]`) : List of class labels
+- `path_to_labels` (`str`) : Path to file with labels. Labels are overrided if it's set.
+- `mean_values` (`list[int | float]`) Normalization values to be subtracted from the image channels during preprocessing.
+- `pad_value` (`int`) Pad value used during the resize resize_image_letterbox operation embedded within the model.
+- `resize_type` (`str`) : The method of resizing the input image. Valid options include `crop`, `standard`, `fit_to_window`, and `fit_to_window_letterbox`.
+- `reverse_input_channels` (`bool`) : Whether to reverse the order of input channels.
+- `scale_values` (`list[int | float]`): Normalization values used to divide the image channels during preprocessing.
+
+## Input format
+
+The input format for action classification tasks differs from other vision tasks due to the nature of video data.
+The input tensor includes additional dimensions to accommodate the video format.
+It's often refered as single alphabet, and each alphabet means as below.
+
+- N : Batch size
+- S : Numer of clips x Number of crops
+- C : Number of channels
+- T : Time
+- H : Height
+- W : Width
+
+The input should be provided as a single clip in THWC format.
+Depending on the specified layout, the input will be transformed into either NSTHWC or NSCTHW format.
+Unlike other vision model wrappers that utilize OpenVINO™'s PrePostProcessors (PPP) for preprocessing,
+the ActionClassificationModel performs its preprocessing due to the current lack of video format support in OpenVINO™ PPP.
+
+## output format
+
+The output is encapsulated in a ClassificationResult object, which includes the indices, labels, and logits of the top predictions.
+At present, saliency maps, feature vectors, and raw scores are not provided.
--- a/model_api/python/docs/ActionClassification.md
+++ b/model_api/python/docs/ActionClassification.md
@@ -10,7 +10,7 @@ Note that it isn't a subclass of `ImageModel`. It gets video as input so it is d
 
 Utilizing the ActionClassificationModel is similar to other model wrappers, with the primary difference being the preparation of video clip inputs instead of single images.
 
-Below is an example demonstrating how to initialize the model with OpenVINO™ IR files and classify actions in a video clip.
+Below is an example demonstrating how to initialize the model with OpenVINO IR files and classify actions in a video clip.
 
 
 ```python
@@ -30,7 +30,7 @@ input_data = np.stack([cap.read()[1] for _ in range(8)])
 # define the path to action classification model in IR format
 model_path = "action_classification.xml"
 
-# create adapter for OpenVINO™ runtime, pass the model path
+# create adapter for OpenVINO runtime, pass the model path
 inference_adapter = OpenvinoAdapter(create_core(), model_path, device="CPU")
 
 # instantiate the ActionClassificationModel wrapper
@@ -70,8 +70,8 @@ It's often refered as single alphabet, and each alphabet means as below.
 
 The input should be provided as a single clip in THWC format.
 Depending on the specified layout, the input will be transformed into either NSTHWC or NSCTHW format.
-Unlike other vision model wrappers that utilize OpenVINO™'s PrePostProcessors (PPP) for preprocessing,
-the ActionClassificationModel performs its preprocessing due to the current lack of video format support in OpenVINO™ PPP.
+Unlike other vision model wrappers that utilize OpenVINO's PrePostProcessors (PPP) for preprocessing,
+the ActionClassificationModel performs its preprocessing due to the current lack of video format support in OpenVINO PPP.
 
 ## output format
 
--- a/model_api/python/docs/ActionClassification.md
+++ b/model_api/python/docs/ActionClassification.md
@@ -22,11 +22,6 @@ from model_api.models import ActionClassificationModel
 from model_api.adapters import OpenvinoAdapter, create_core
 
 
-# load video and make a clip as input
-cap = cv2.VideoCapture("sample.mp4")
-input_data = np.stack([cap.read()[1] for _ in range(8)])
-
-
 # define the path to action classification model in IR format
 model_path = "action_classification.xml"
 
@@ -36,6 +31,10 @@ inference_adapter = OpenvinoAdapter(create_core(), model_path, device="CPU")
 # instantiate the ActionClassificationModel wrapper
 # setting preload=True loads the model onto the CPU within the adapter0
 action_cls_model = ActionClassificationModel(inference_adapter, preload=True)
+
+# load video and make a clip as input
+cap = cv2.VideoCapture("sample.mp4")
+input_data = np.stack([cap.read()[1] for _ in range(action_cls_model.clip_size)])
 
 # perform preprocessing, inference, and postprocessing
 results = action_cls_model(input_data)
--- a/model_api/python/model_api/adapters/utils.py
+++ b/model_api/python/model_api/adapters/utils.py
@@ -527,5 +527,5 @@ class InputTransform:
         if self.is_trivial:
             return inputs
         if self.reverse_input_channels:
-            inputs = cv2.cvtColor(inputs, cv2.COLOR_BGR2RGB)
+            inputs = inputs[...,::-1]
         return (inputs - self.means) / self.std_scales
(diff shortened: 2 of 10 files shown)


model_api/python/model_api/adapters/utils.py
```diff
@@ -527,5 +529,5 @@ def __call__(self, inputs):
         if self.is_trivial:
             return inputs
         if self.reverse_input_channels:
-            inputs = cv2.cvtColor(inputs, cv2.COLOR_BGR2RGB)
+            inputs = inputs[...,::-1]
```
sovrasov, line 532: I'd use cv2 for images (perhaps it's faster)

model_api/python/model_api/adapters/utils.py
```diff
@@ -41,6 +41,8 @@ def from_shape(shape):
             return "CHW" if shape[0] in range(1, 5) else "HWC"
         if len(shape) == 4:
             return "NCHW" if shape[1] in range(1, 5) else "NHWC"
+        if len(shape) == 6:
+            return "?NCTHW" if shape[1] in range(1, 5) else "?NTHWC"
```
sovrasov, line 45: What does `?` mean?

model_api/python/model_api/models/action_classification.py
```diff
@@ -0,0 +1,194 @@
...
+
+        Also, it keeps the size of original image and resized one as `original_shape` and `resized_shape`
+        in the metadata dictionary.
+
+        Note:
+            It supports only models with single image input. If the model has more image inputs or has
+            additional supported inputs, the `preprocess` should be overloaded in a specific wrapper.
+
+        Args:
+            inputs (ndarray): a single image as 3D array in HWC layout
```
sovrasov, line 155: This description seems to be outdated

model_api/python/model_api/models/action_classification.py
```diff
@@ -0,0 +1,194 @@
...
+    def postprocess(self, outputs: dict[str, np.ndarray], meta: dict[str, Any]) -> ClassificationResult:
+        """Post-process."""
+        logits = next(iter(outputs.values())).squeeze()
+        return get_multiclass_predictions(logits)
+
+
+def get_multiclass_predictions(logits: np.ndarray) -> ClassificationResult:
+    """Get multiclass predictions."""
+    index = np.argmax(logits)
+    return ClassificationResult([(index, index, logits[index])], np.ndarray(0), np.ndarray(0), np.ndarray(0))
```
sovrasov, line 194: Should there be also a string label?

tests/python/accuracy/public_scope.json
```diff
@@ -395,5 +395,16 @@
                 ]
             }
         ]
+
+    },
+        {
+        "name": "otx_models/action_cls_xd3_kinetic.xml",
```
sovrasov, line 401: Please modify tests/python/accuracy/prepare_data.py accordingly to get this file downloaded

model_api/python/model_api/adapters/utils.py
eunwoosh: mmaction2 adds one more dimension and it's called `N_crops` in doc string. but I don't know what it is exactly. There is no explanation about that.

model_api/python/model_api/adapters/utils.py
eunwoosh: I changed it for processing staked images because previous code can only single image at a time.
Are there other reason to use cv2 I missed?

model_api/python/model_api/models/action_classification.py
eunwoosh: I missed it. Thanks.

model_api/python/model_api/adapters/utils.py
eunwoosh: I dicussed it w/ Jaeguk and we understand what it is. I updated it.

model_api/python/model_api/adapters/utils.py
sovrasov: That's performance, we don't know if numpy is faster or slower

model_api/python/model_api/models/action_classification.py
```diff
@@ -0,0 +1,220 @@
...
+            configuration (dict, optional): it contains values for parameters accepted by specific
+              wrapper (`labels` `mean_values`, etc.) which are set as data attributes
+            preload (bool, optional): a flag whether the model is loaded to device while
+              initialization. If `preload=False`, the model must be loaded via `load` method before inference
+
+        Raises:
+            WrapperError: if the wrapper failed to define appropriate inputs for images
+        """
+        super().__init__(inference_adapter, configuration, preload)
+        self.image_blob_names, self.image_info_blob_names = self._get_inputs()
```
sovrasov, line 75: Is `self.image_info_blob_names` used somewhere?

model_api/python/model_api/models/action_classification.py
```diff
@@ -0,0 +1,220 @@
...
+from .utils import ClassificationResult, load_labels
+
+if TYPE_CHECKING:
+    from model_api.adapters.inference_adapter import InferenceAdapter
+
+
+class ActionClassificationModel(Model):
+    """An wrapper for an action classification model
+
+    The ActionClassificationModel has 1 or more inputs with images - 6D tensors.
```
sovrasov, line 35: Is that true from wrapper user's perspective?

model_api/python/model_api/models/action_classification.py
```diff
@@ -0,0 +1,220 @@
...
+
+from .model import Model
+from .types import BooleanValue, ListValue, NumericalValue, StringValue
+from .utils import ClassificationResult, load_labels
+
+if TYPE_CHECKING:
+    from model_api.adapters.inference_adapter import InferenceAdapter
+
+
+class ActionClassificationModel(Model):
```
sovrasov, line 32: Overall, I thinks we need also a method which can tell users which is the optimal clips size. If user passes a shorter clip, it can be extended with empty frames.
Otherwise, the model is going to have dynamic input shape, which is not intended I believe

model_api/python/docs/ActionClassification.md
```diff
@@ -0,0 +1,79 @@
...
+import numpy as np
+# import model wrapper class
+from model_api.models import ActionClassificationModel
+# import inference adapter and helper for runtime setup
+from model_api.adapters import OpenvinoAdapter, create_core
+
+
+# load video and make a clip as input
+cap = cv2.VideoCapture("sample.mp4")
+input_data = np.stack([cap.read()[1] for _ in range(8)])
```
sovrasov, line 27: Do we always have to use 8 frames per clip?

model_api/python/model_api/models/action_classification.py
eunwoosh: It isn't. I'll remove it.

model_api/python/model_api/models/action_classification.py
```diff
@@ -0,0 +1,220 @@
...
+    """An wrapper for an action classification model
+
+    The ActionClassificationModel has 1 or more inputs with images - 6D tensors.
+    It may support additional inputs - 4D tensors.
+    Two input formats are supported. One is 'NSCTHW' and another one is 'NSTHWC'.
+    What each alphabet means are as below.
```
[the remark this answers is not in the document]
eunwoosh: done https://github.com/openvinotoolkit/model_api/pull/178/commits/7e4dd83d40c20d88ab7dbe19f42315cb2e21dc2d

model_api/python/model_api/models/action_classification.py
```diff
@@ -0,0 +1,220 @@
...
+from .model import Model
+from .types import BooleanValue, ListValue, NumericalValue, StringValue
+from .utils import ClassificationResult, load_labels
+
+if TYPE_CHECKING:
+    from model_api.adapters.inference_adapter import InferenceAdapter
+
+
+class ActionClassificationModel(Model):
+    """An wrapper for an action classification model
```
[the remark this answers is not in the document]
eunwoosh: done https://github.com/openvinotoolkit/model_api/pull/178/commits/7e4dd83d40c20d88ab7dbe19f42315cb2e21dc2d

model_api/python/model_api/models/action_classification.py
```diff
@@ -0,0 +1,220 @@
...
+
+    The ActionClassificationModel has 1 or more inputs with images - 6D tensors.
+    It may support additional inputs - 4D tensors.
+    Two input formats are supported. One is 'NSCTHW' and another one is 'NSTHWC'.
+    What each alphabet means are as below.
+    N => batch size / S => number of clips x number of crops / C => number of channels
+    T => time / H => height / W => width
+
+    Video format is different from image format, so OpenVINO PrePostProcessors isn't available.
+    For that reason, basic preprocessing such as resize and normalize are conducted in this class.
```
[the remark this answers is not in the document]
eunwoosh: done https://github.com/openvinotoolkit/model_api/pull/178/commits/7e4dd83d40c20d88ab7dbe19f42315cb2e21dc2d

model_api/python/model_api/models/action_classification.py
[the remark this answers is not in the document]
eunwoosh: done https://github.com/openvinotoolkit/model_api/pull/178/commits/7e4dd83d40c20d88ab7dbe19f42315cb2e21dc2d

model_api/python/model_api/models/action_classification.py
eunwoosh: done https://github.com/openvinotoolkit/model_api/pull/178/commits/7e4dd83d40c20d88ab7dbe19f42315cb2e21dc2d

model_api/python/model_api/models/action_classification.py
eunwoosh: I found that this class assumes input should be single video clip w/ 4D tensors. I updated it https://github.com/openvinotoolkit/model_api/pull/178/commits/7e4dd83d40c20d88ab7dbe19f42315cb2e21dc2d

model_api/python/docs/ActionClassification.md
eunwoosh: no it depends on user's configuration. In real scenario, user needs to sample # clips from video.

model_api/python/model_api/models/action_classification.py
eunwoosh: I think it's better to raise an error which tells "need to set dynamic input shape" when user gives inputs w/ various shapes although user gives OV IR w/ fixed input shape.

model_api/python/model_api/adapters/utils.py
eunwoosh: I tried to compare speed using code below
```
import cv2
import numpy as np
import time

IMAGE_PATH = "data/coco/train2017/000000149165.jpg"

def run_cv(img: np.array):
    res = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
    return res

def run_np(img: np.array):
    res = img[..., ::-1]
    return res

if __name__ == "__main__":
    img = cv2.imread(IMAGE_PATH)

    start_time = time.perf_counter()
    for _ in range(10000000):
        # run_np(img)
        run_cv(img)
    print(time.perf_counter() - start_time)
    
```
and result is 

- cv : 1.1237245709635317
- numpy : 0.017020812025293708

It seems that numpy is faster than cv2. How about using numpy?

sovrasov: Could you re-run the test for a bigger resolution to avoid the case when numpy does just an index remapping without changing the memory layout? The time should grow linearly with the amount memory to restructure.

model_api/python/model_api/models/action_classification.py
sovrasov: I'm ok with that. OV's error in that case could be not quite verbose

model_api/python/docs/ActionClassification.md
sovrasov: Right, but what is the #? In the real application users could run several different IRs with that wrapper, each of the IRs could have a different # hardcoded as a static shape.
I propose to add 'getClipSize() -> int' method to the wrapper. It should read # from the static input shape and return it. If the input shape is dynamic, the method could return -1 (as a sign of dynamic shape in OV).

eunwoosh: \# means num_clips user set. Actually num_clips is already stored as an attribute here
```python
class ActionClassificationModel(Model):
    def __init__(...):
        ...
        if self.nscthw_layout:
            self.n, self.s, self.c, self.t, self.h, self.w = self.inputs[
                self.image_blob_name
            ].shape
        else:
            self.n, self.s, self.t, self.h, self.w, self.c = self.inputs[
                self.image_blob_name
            ].shape
``` 
`self.t` is num_clips.

model_api/python/model_api/adapters/utils.py
eunwoosh: I don't know the reason, but speed changes a lot so I tested original resolution once again w/ code below.
```Python
import cv2
import numpy as np
import time

IMAGE_PATH = "data/coco/train2017/000000149165.jpg"

def run_cv(img: np.array):
    res = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
    return res

def run_np(img: np.array):
    res = img[..., ::-1]
    return res

if __name__ == "__main__":
    img = cv2.imread(IMAGE_PATH)
    # img = cv2.resize(img, (4096, 4096))

    start_time = time.perf_counter()
    for _ in range(10000):
        run_np(img)
        # run_cv(img)
    print(time.perf_counter() - start_time)
```


Here is the result w/ (612, 612, 3)

- cv : 0.119686498073861
- numpy : 0.0037762820720672607

these are result w/ (4096, 4096, 3)

- cv : 295.947567245923
- numpy : 0.119686498073861

model_api/python/docs/ActionClassification.md
sovrasov: Great, then that'd be easy to implement the method & mention it in the documentation

model_api/python/model_api/adapters/utils.py
sovrasov: I see, that's because numpy actually uses mapping rather than true reordering.
Please, use `res = np.ascontiguousarray(img[..., ::-1])`

eunwoosh: I checked that it spends 555.3918372320477 seconds. It's slower than cv2 if forcing reordering.
I don't know well about this part, but is it necessary to force actual reordering?

eunwoosh: reverted it. https://github.com/openvinotoolkit/model_api/pull/178/commits/4708d2bcbd61e6d1e911161132af885cbdab210f

model_api/python/docs/ActionClassification.md
eunwoosh: done https://github.com/openvinotoolkit/model_api/pull/178/commits/4708d2bcbd61e6d1e911161132af885cbdab210f

PR #181, opened 2024-06-11
Add visual prompting wrappers

sovrasov:
Only python implementation for now.
- SAM encoder/decoder are implemented as Model subclasses
- Prompters are standalone entities that operate by encoder/decoder
- No prompter parameters are exposed to user, because there is no experiments demonstrating their impact available for review.

Usage sample / docs come in the future PR


model_api/python/model_api/models/visual_prompting.py
```diff
@@ -0,0 +1,675 @@
...
+                    prediction[self.decoder_model.output_blob_name]
+                )
+                used_points[label].append(points_score)
+
+        # check overlapping area between different label masks
+        _inspect_overlapping_areas(predicted_masks, used_points)
+
+        return (predicted_masks, used_points)
+
+    def reset_reference_info(self) -> None:
```
maxxgx, line 291: Since there is a property called `reference_features`, it would make sense to also rename this method `reset _reference_features`.

model_api/python/model_api/models/visual_prompting.py
```diff
@@ -0,0 +1,675 @@
...
+
+        raise RuntimeError("Reference features are not generated")
+
+    def learn(
+        self,
+        image: np.ndarray,
+        boxes: Prompt | None = None,
+        points: Prompt | None = None,
+        reset_features: bool = False
+    ) -> tuple[VisualPromptingFeatures, np.ndarray]:
```
maxxgx, line 139: How is learning on multiple images/labels handled?

model_api/python/model_api/models/visual_prompting.py
```diff
@@ -0,0 +1,675 @@
...
+        self.decoder_model = decoder_model
+
+    def infer(
+        self,
+        image: np.ndarray,
+        boxes: Prompt | None = None,
+        points: Prompt | None = None,
+    ) -> VisualPromptingResult:
+        if boxes is None and points is None:
+            raise RuntimeError("boxes or points prompts are required for inference")
```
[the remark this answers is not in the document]
sovrasov: Boxes and points can be used together. Each prompt is evaluated (a point or a box) and then it contributes to the corresponding reference feature vector. Corresponding here means that reference vectors are grouped by labels.

model_api/python/model_api/models/visual_prompting.py
sovrasov: Zero-shot learning is done on each image individually. Learn() returns the reference features vectors for each label. Labels are contained in prompts. Later, the reference feature vectors can be used to obtain object masks on other images.

Currently, learning the same classes on multiple images is not considered, since first we need to prove that feature vector arithmetic somehow works in that scenario.
However, consequent learn() calls still make sense if we pass new prompts with unseen labels. In that case, a list of reference features would be extended with the ones corresponding to new labels.

model_api/python/model_api/models/visual_prompting.py
sovrasov: Updated

model_api/python/model_api/models/visual_prompting.py
```diff
@@ -0,0 +1,757 @@
...
+
+
+class VisualPromptingFeatures(NamedTuple):
+    feature_vectors: np.ndarray
+    used_indices: np.ndarray
+
+
+class Prompt(NamedTuple):
+    data: list[np.ndarray] | np.ndarray
+    labels: list[np.ndarray | int] | np.ndarray
```
maxxgx, line 31: `Prompt` class seems to contain data related to multiple prompts. I think it's cleaner to have it be a single prompt and then use a list. Something like:
```python
class Prompt(NamedTuple):
    data: np.ndarray
    label: int

...

def infer(
    self,
    image: np.ndarray,
    boxes: list[Prompt] | None = None,
    points: list[Prompt] | None = None,
) -> VisualPromptingResult:
```

model_api/python/model_api/models/visual_prompting.py
```diff
@@ -0,0 +1,757 @@
...
+                inputs_decoder.pop("label")
+                if "point_coords" in inputs_decoder:
+                    # bboxes and points
+                    inputs_decoder["image_embeddings"] = image_embeddings
+                    prediction = self._predict_masks(
+                        inputs_decoder, original_shape, is_cascade=self._is_cascade
+                    )
+                    masks = prediction["upscaled_masks"]
+                else:
+                    raise RuntimeError("Prompts other than points are not supported")
```
sungchul2, line 237: @maxxgx Do you consider a case with polygons (mask or scribble generated from Geti)?
If so, we can refer this implementation I did for demo last year.
https://github.com/sungchul2/training_extensions/blob/a5f869a558f90afda500495cde5431679be2f023/src/otx/algorithms/visual_prompting/adapters/torch/processor/zero_shot_processor.py#L576-L579

maxxgx: Indeed, we also need polygon type prompts. Thanks for the reminder!

sovrasov: @sungchul2 could you prepare a slide with requirements for polygon annotation and send it to me, so we can discuss it?

model_api/python/model_api/models/visual_prompting.py
sovrasov: Done

model_api/python/model_api/models/utils.py
```diff
@@ -136,6 +136,61 @@ def __str__(self):
...
+                obj_str += "] "
+
+        return obj_str.strip()
+
+
+class ZSLVisualPromptingResult(NamedTuple):
+    data: dict[int, PredictedMask]
+
+    def __str__(self) -> str:
+        return ", ".join(str(self.data[k]) for k in self.data)
```
maxxgx, line 191: I think `data` is a bit too generic here, it would be more user-friendly to either implement a `get_mask(label: int)` or `__getitem__` methods.

model_api/python/model_api/models/sam_models.py
```diff
@@ -0,0 +1,204 @@
...
+            outputs (dict[str, np.ndarray]): The output of the model.
+            meta (dict[str, Any]): Contain label and original size.
+
+        Returns:
+            (dict[str, np.ndarray]): The postprocessed output of the model.
+        """
+        probability = np.clip(outputs["scores"], 0.0, 1.0)
+        hard_prediction = (
+            outputs[self.output_blob_name].squeeze(0) > self.mask_threshold
+        ).astype(np.uint8)
```
Wovchena, line 198:
```suggestion
        )
```
`soft_prediction = hard_prediction * probability.reshape(-1, 1, 1)` is fine masking with bool. OpenVINO Tensors also support bool: https://docs.openvino.ai/2022.3/api/ie_python_api/_autosummary/openvino.runtime.Type.html

Wovchena (approved):

model_api/python/model_api/models/sam_models.py
sovrasov: Updated

model_api/python/model_api/models/utils.py
sovrasov: Added

model_api/python/model_api/models/visual_prompting.py
```diff
@@ -0,0 +1,769 @@
...
+    ):
+        self.encoder = encoder_model
+        self.decoder = decoder_model
+
+    def infer(
+        self,
+        image: np.ndarray,
+        boxes: list[Prompt] | None = None,
+        points: list[Prompt] | None = None,
+        polygons: list[Prompt] | None = None,
```
sungchul2, line 61: Basic SAM inference doesn't need `polygons`.
If we already have `polygons`, we don't need to run `infer`.

sovrasov: Deleted

sovrasov:
@sungchul2 please have a look at polygons support for ZSL

model_api/python/model_api/models/visual_prompting.py
```diff
@@ -0,0 +1,784 @@
...
+                    "bboxes": [box.label for box in boxes] if boxes else None,
+                    "points": [point.label for point in points] if points else None,
+                },
+                "orig_size": image.shape[:2],
+            },
+        )
+
+        if polygons is not None:
+            for poly in polygons:
+                processed_prompts.append({"polygon": poly.data, "label": poly.label})
```
sungchul2, line 230: I think the additional workflow is needed to avoid getting duplicated masks:
check if there is bbox or point prompt that targets the same object with polygon -> if so, remove corresponding bbox or point
What do you think?

sovrasov: I think we should discuss this with Max. I'm not sure if prompts cleaning should be a responsibility of the ZSL pipeline, since the annotation engine is somewhat external.

maxxgx: What happens if we have duplicate prompts? And if they have different labels?

sungchul2: > I think we should discuss this with Max. I'm not sure if prompts cleaning should be a responsibility of the ZSL pipeline, since the annotation engine is somewhat external.

Ah, I agree with your opinion. Let's pass it.

sungchul2: > What happens if we have duplicate prompts? And if they have different labels?

No problem because `learn` just outputs reference features corresponding to given prompts.
But there are reference features that target the same objects and duplicated masks will be generated.
For this case, `_inspect_overlapping_areas` will filter those that have the same label.
https://github.com/sovrasov/model_api/blob/d6c9ad0a0ea24b9033c3d7637e239c27bba473fe/model_api/python/model_api/models/visual_prompting.py#L737-L741
But masks with different labels will not be applicable, resulting in duplicate results.

maxxgx: Thanks for the explanation. I agree that it's user's responsibility for processing the prompts.

PR #182, opened 2024-06-11
Drop python 3.8 support

sovrasov:
Python 3.8 is EOL is 31 Oct 2024: https://endoflife.date/python

### DIFF
--- a/.github/workflows/test_precommit.yml
+++ b/.github/workflows/test_precommit.yml
@@ -135,13 +135,14 @@ jobs:
         mkdir build && cd build
         MSYS_NO_PATHCONV=1 cmake ../tests/cpp/precommit/ -DOpenVINO_DIR=$GITHUB_WORKSPACE/w_openvino_toolkit_windows_2024.1.0.15008.f4afc983258_x86_64/runtime/cmake -DOpenCV_DIR=$GITHUB_WORKSPACE/opencv/opencv/build -DCMAKE_CXX_FLAGS=/WX
         cmake --build . --config Release -j $((`nproc`*2+2))
-    - name: Run test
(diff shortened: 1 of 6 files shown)


.github/workflows/test_precommit.yml
```diff
@@ -148,7 +148,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04, ubuntu-22.04, macos-12]
-        python-version: [3.8, 3.9, '3.10', '3.11']
+        python-version: [3.9, '3.10', '3.11']
```
Wovchena, line 151: Can 3.8 be kept here and in `model_api/python/pyproject.toml`? It's not costly to keep the support

sovrasov: Yes, that's possible, but my intention was to ease type annotations in new PRs.
However, OV 2024.1 still supports 3.8, probably they'll wait till October, and then drop it. Let's keep 3.8 until that point then

Wovchena: Any feature would be a stronger reason to drop 3.8 compared to just keeping the support. You are the only users of model_api. So if you don't need 3.8, feel free to drop now.

Wovchena (approved):

sovrasov:
@Wovchena I had to replace sanity tests on Win to running a sample, because of somewhat similar to https://github.com/google/googletest/issues/2628
The problem occurs even though I updated gtest commit to the latest version.

sovrasov:
Apparently the gtest problem is still valid: https://github.com/google/googletest/issues/4556

Wovchena (approved):

PR #184, opened 2024-06-12
Revert changes required to pass CI after python version change

sovrasov:
- Restore pip cache
- Del forcing cpp14 for win tests

### DIFF
--- a/.github/workflows/test_precommit.yml
+++ b/.github/workflows/test_precommit.yml
@@ -78,6 +78,7 @@ jobs:
     - uses: actions/setup-python@v4
       with:
         python-version: 3.9
+        cache: pip
     - name: Create and start a virtual environment
       run: |
         python -m venv venv
@@ -108,7 +109,7 @@ jobs:
     - uses: actions/setup-python@v4
       with:
         python-version: 3.9
-        #cache: pip
+        cache: pip
     - name: Create and start a virtual environment
       shell: bash
       run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/test_precommit.yml
```diff
@@ -108,7 +109,7 @@ jobs:
     - uses: actions/setup-python@v4
       with:
         python-version: 3.9
-        #cache: pip
+        cache: pip
```
Wovchena, line 112: Just in case, I don't think we need caching at all. Enabling the cache just downloads an archive from GitHub instead of PyPI. Moreover as far as I understand, modern pip downloads packages asynchronously all at the same time which may even outperform archive download.

sovrasov: Haven't heard pip is parallel, I found a comparison with other python package installers here: https://pythonspeed.com/articles/faster-pip-installs/
pip logs also make an impression the packages are downloaded installed sequentially.
pip is faster when cache is not empty, according to the paper. I'd expect, GHA could fetch it's own cache faster than pip downloads packages from some 3rd party mirrors

sovrasov: https://github.com/pypa/pip/issues/825

Wovchena: Ok. Up to you

Wovchena (approved):

PR #185, opened 2024-06-21
[cpp] Fix num classes check for cls models with XAI

sovrasov:
# What does this PR do?

Cpp classification wrapper now enforces labels consistency check in constructor on models with XAI outputs.

### DIFF
--- a/model_api/cpp/models/src/classification_model.cpp
+++ b/model_api/cpp/models/src/classification_model.cpp
@@ -158,6 +158,25 @@ std::vector<std::string> get_non_xai_names(const std::vector<ov::Output<ov::Node
     return outputNames;
 }
 
+std::vector<size_t> get_non_xai_output_indices(const std::vector<ov::Output<ov::Node>>& outputs) {
+    std::vector<size_t> outputIndices;
+    outputIndices.reserve(std::max(1, int(outputs.size()) - 2));
+    size_t idx = 0;
+    for (const ov::Output<ov::Node>& output : outputs) {
+        if (output.get_names().count(saliency_map_name) > 0) {
+            continue;
+        }
+        else if (output.get_names().count(feature_vector_name) > 0) {
+            continue;
+        }
+        else {
+            outputIndices.push_back(idx);
+        }
+        ++idx;
+    }
+    return outputIndices;
+}
+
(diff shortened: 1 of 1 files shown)


model_api/cpp/models/src/classification_model.cpp
```diff
@@ -480,18 +500,15 @@ void ClassificationModel::prepareInputsOutputs(std::shared_ptr<ov::Model>& model
         if (outputShape.size() == 4 && (outputShape[ov::layout::height_idx(outputLayout)] != 1 ||
                                         outputShape[ov::layout::width_idx(outputLayout)] != 1)) {
             throw std::logic_error("Classification model wrapper supports topologies only"
-                                " with 4-dimensional output which has last two dimensions of size 1");
+                                " with 4-dimensional output hich has last two dimensions of size 1");
```
Wovchena, line 503:
```suggestion
                                " with 4-dimensional output which has last two dimensions of size 1");
```

model_api/cpp/models/src/classification_model.cpp
```diff
@@ -480,18 +500,15 @@ void ClassificationModel::prepareInputsOutputs(std::shared_ptr<ov::Model>& model
...
 
         size_t classesNum = outputShape[ov::layout::channels_idx(outputLayout)];
         if (topk > classesNum) {
             throw std::logic_error("The model provides " + std::to_string(classesNum) + " classes, but " +
                                 std::to_string(topk) + " labels are requested to be predicted");
         }
-        if (classesNum == labels.size() + 1) {
-            labels.insert(labels.begin(), "other");
-            slog::warn << "Inserted 'other' label as first." << slog::endl;
-        } else if (classesNum != labels.size()) {
```
Wovchena, line 494: This was the only place adding `other`. I think you've removed `other` for all models, not just for xai,

Wovchena (changes requested):

model_api/cpp/models/src/classification_model.cpp
sovrasov: That's right, I had to eliminate this ad-hoc heuristic completely.
Inserting `other` to the labels list requires some prior knowledge of the actual insertion place, which I'm not aware of, at least for OTX/Geti usecases.

Wovchena (approved):

PR #191, opened 2024-08-22
Fix MRCNN label index out of bound

eugene123tw:
# What does this PR do?

Fixes MRCNN fix label index out of bound

## Before submitting

### DIFF
--- a/model_api/python/model_api/models/instance_segmentation.py
+++ b/model_api/python/model_api/models/instance_segmentation.py
@@ -156,10 +156,6 @@ class MaskRCNNModel(ImageModel):
         masks = outputs[self.output_blob_name["masks"]]
         if not self.is_segmentoly:
             labels += 1
-        if not self.labels:
-            str_labels = (f"#{label}" for label in labels)
-        else:
-            str_labels = (self.labels[label] for label in labels)
 
         inputImgWidth, inputImgHeight = (
             meta["original_shape"][1],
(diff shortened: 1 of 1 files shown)


model_api/python/model_api/models/instance_segmentation.py
```diff
@@ -199,14 +195,17 @@ def postprocess(self, outputs, meta):
...
+        for box, confidence, cls, raw_mask in zip(boxes, scores, labels, masks):
             x1, y1, x2, y2 = box
             if (x2 - x1) * (y2 - y1) < 1 or (
                 confidence <= self.confidence_threshold and not has_feature_vector_name
             ):
                 continue
+            str_label = (
+                self.labels[cls]
+                if self.labels and cls < len(self.labels)
+                else f"#{cls}"
```
sovrasov, line 207: Why do we need to generate predictions which are not aligned with the labels we have in rt_info?

eugene123tw: @sovrasov We could skip it by

```py
for box, confidence, cls, raw_mask in zip(boxes, scores, labels, masks):
    x1, y1, x2, y2 = box
    area = (x2 - x1) * (y2 - y1)

    if area < 1 or (confidence <= self.confidence_threshold and not has_feature_vector_name):
        continue

    # Skip if label index is out of bounds
    if self.labels and cls >= len(self.labels):
        continue

    # Get label string
    str_label = self.labels[cls] if self.labels else f"#{cls}"
```

sovrasov: That looks better, and would allow us to prevent assigning undefined labels

sovrasov (approved):

PR #195, opened 2024-09-12
[python] Add batch inference emulation 

sovrasov:
Batch inference emulation is done by enqueuing batch items for async inference.

That design allows to avoid usage of AsyncPipeline, simplifying batch processing for end users.
However, if obtaining every batch element is somehow expensive, AsyncPipeline allows to squeeze more performance by running infer requests in parallel to preparation of the next batch elements.

### DIFF
--- a/model_api/python/model_api/models/model.py
+++ b/model_api/python/model_api/models/model.py
@@ -14,6 +14,7 @@
  limitations under the License.
 """
 
+from contextlib import contextmanager
 import logging as log
 import re
 
@@ -97,6 +98,7 @@ class Model:
         self.model_loaded = False
         if preload:
             self.load()
+        self.callback_fn = lambda _: None
 
     def get_model(self):
         """Returns the ov.Model object stored in the InferenceAdapter.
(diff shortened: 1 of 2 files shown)


model_api/python/model_api/models/model.py
```diff
@@ -402,6 +404,40 @@ def __call__(self, inputs):
...
+
+        Returns:
+            list: a list of inference results
+        """
+        self.await_all()
+
+        completed_results = {}
+
+        @contextmanager
+        def tmp_callback(*args, **kwds):
```
eunwoosh, line 422:
```suggestion
        def tmp_callback():
```

model_api/python/model_api/models/model.py
```diff
@@ -402,6 +404,40 @@ def __call__(self, inputs):
...
+            old_callback = self.callback_fn
```
eunwoosh, line 423: If user set own callbacks, then it's ignored. Is it ok?

eunwoosh (approved):
LGTM. I have minor comments. Please ignore it if you need :)

model_api/python/model_api/models/model.py
sovrasov: This functionality saves user's callback, and then restores it

model_api/python/model_api/models/model.py
sovrasov: Fixed

model_api/python/model_api/models/model.py
sovrasov: User's callback is not supposed to be executed during batch inference, so we have to backup it for the future

PR #198, opened 2024-09-17
[python] Fix classification usage via OVMS

sovrasov:
3rd party and preprocessed models should be discriminated in cls wrapper

### DIFF
--- a/model_api/python/model_api/models/classification.py
+++ b/model_api/python/model_api/models/classification.py
@@ -80,11 +80,13 @@ class ClassificationModel(ImageModel):
             if self.output_raw_scores:
                 self.out_layer_names.append(self.raw_scores_name)
         except (RuntimeError, AttributeError):
+            # exception means we have a non-ov model, or a model behind OVMS
+            # with already inserted softmax and topk
             if self.embedded_processing and len(self.outputs) >= 2:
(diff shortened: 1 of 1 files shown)


model_api/python/model_api/models/classification.py
```diff
@@ -80,9 +80,14 @@ def __init__(self, inference_adapter, configuration=dict(), preload=False):
             if self.output_raw_scores:
                 self.out_layer_names.append(self.raw_scores_name)
         except (RuntimeError, AttributeError):
-            self.embedded_topk = False
-            self.out_layer_names = _get_non_xai_names(self.outputs.keys())
-            self.raw_scores_name = self.out_layer_names[0]
+            if self.embedded_processing and len(self.outputs) >= 2:
```
eunwoosh, line 83: Is it for OVMS use case? It's a little bit hard to understand what it's for to me. How about adding a docstring for explanation?

sovrasov: It's for the case of already preprocessed models behind the OVMS. We can't check the model nodes, and get an exception. But we actually know the exact outputs names and can use them right away.

eunwoosh (approved):

PR #200, opened 2024-09-19
Fix masks merging in ZSL VPT

sovrasov:
[] operation was expecting a bool array


model_api/python/model_api/models/visual_prompting.py
```diff
@@ -286,7 +286,7 @@ def learn(
                     masks = _polygon_to_mask(inputs_decoder["polygon"], *original_shape)
                 else:
                     raise RuntimeError("Unsupported type of prompt")
-                ref_mask[masks] += 1
+                ref_mask = np.where(masks, 1, ref_mask)
             ref_mask = np.clip(ref_mask, 0, 1)
```
eunwoosh, line 290:
```suggestion
```
Isn't it ok to remove?

sovrasov: Done

maxxgx (approved):
I tested the change, and it fixes learning on polygon prompts. Thanks!

PR #201, opened 2024-09-24
Tiling for semantic segmentation

sovrasov:
# What does this PR do?

Adds semantic segmentation tiler and tests

### DIFF
--- a/model_api/cpp/tilers/src/semantic_segmentation.cpp
+++ b/model_api/cpp/tilers/src/semantic_segmentation.cpp
@@ -1,0 +1,100 @@
+/*
+// Copyright (C) 2024 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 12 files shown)


model_api/cpp/tilers/src/semantic_segmentation.cpp
```diff
@@ -0,0 +1,100 @@
...
+    const int num_classes = soft_prediction.channels();
+    const size_t step_rows = soft_prediction.step[0] / sizeof(float);
+    const size_t step_cols = soft_prediction.step[1] / sizeof(float);
+
+    for (int y = 0; y < soft_prediction.rows; ++y) {
+        for (int x = 0; x < soft_prediction.cols; ++x) {
+            int weight = normalize_factor.at<int>(y, x);
+            if (weight > 0) {
+                for (int c = 0; c < num_classes; ++c) {
+                    data[y * step_rows + x * step_cols + c] /= weight;
```
eugene123tw, line 38: I'm curious, is it possible to use `cv::divide` or other array operations? Syntax-wise, it looks cleaner, but I'm not sure if it's more complex than it seems in OpenCV. If it is, the current approach works fine too.

sovrasov: I agree `cv::divide` or even `/` operation looks cleaner, but OpenCV can't do it when the number of channels doesn't match:
```
unknown file: Failure
C++ exception with description "OpenCV(4.5.4) ./modules/core/src/arithm.cpp:647: error: (-209:Sizes of input arguments do not match) The operation is neither 'array op array' (where arrays have the same size and the same number of channels), nor 'array op scalar', nor 'scalar op array' in function 'arithm_op'
" thrown in the test body.
```

PR #203, opened 2024-10-09
Make ZSL VPT masks refinement configurable

sovrasov:
# What does this PR do?

In some cases, it's beneficial to disable extra refinement.

Results on 3-class COCO:

Method | Score (IoU) | Time
-- | -- | --
FastSAM + 3 iters refine | 0.4190 | 01:52
FastSAM + 1 iter refine | 0.4403 | 01:11

### DIFF
--- a/model_api/python/model_api/models/visual_prompting.py
+++ b/model_api/python/model_api/models/visual_prompting.py
@@ -146,6 +146,7 @@ class SAMLearnableVisualPrompter:
         decoder_model: SAMDecoder,
         reference_features: VisualPromptingFeatures | None = None,
         threshold: float = 0.65,
+        masks_extra_refinement: bool = False,
     ):
         """
         Initializes ZSL pipeline.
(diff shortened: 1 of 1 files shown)


model_api/python/model_api/models/visual_prompting.py
```diff
@@ -157,6 +158,7 @@ def __init__(
              Once the features are passed, one can skip learn() method, and start predicting masks right away. Defaults to None.
             threshold (float, optional): Threshold to match vs reference features on infer(). Greater value means a
             stricter matching. Defaults to 0.65.
+            masks_extra_refinement (bool, optional): Flag controlling additional refinement stage on inference.
```
maxxgx, line 161: I suggest passing this as an argument to `infer()` rather than in the constructor, as it seems only relevant there. Also, consider renaming it to `apply_masks_refinement` for clarity. 

It would be helpful to add more context to the docstring, briefly describing the effect of enabling or disabling this option.

sovrasov: Thanks for suggestion, I've updated the implementation

maxxgx (approved):
LGTM, thanks!

The description has performance with FastSAM, shouldn't it be MobileSAM?

PR #207, opened 2024-10-21
Add detection filtering for SSD wrapper

sovrasov:
# What does this PR do?

- Filter now can handle +-inf values
- Boxes with area < 1 pix are not considered, like in iseg wrapper

### DIFF
--- a/model_api/python/model_api/models/ssd.py
+++ b/model_api/python/model_api/models/ssd.py
@@ -75,8 +75,7 @@ class SSD(DetectionModel):
         self.raise_error("Unsupported model outputs")
 
     def _parse_outputs(self, outputs):
-        detections = self.output_parser(outputs)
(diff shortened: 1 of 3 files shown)


model_api/python/model_api/models/ssd.py
```diff
@@ -176,5 +176,6 @@ def __call__(self, outputs):
         return detections
 
 
+_bbox_area_threshold = 1.0
 _saliency_map_name = "saliency_map"
 _feature_vector_name = "feature_vector"
```
eunwoosh, line 181: How about setting these values as attribute or class attirbute?

sovrasov: This will be discussed during MAPI refactoring round

eunwoosh (approved):

PR #208, opened 2024-10-24
Add documentation

ashwinvaidya17:
# What does this PR do?

- Add initial documentation generation scaffold. It sets-up the documentation build that combines C++ and Python docs

## TODO

- [x] Add github actions file
- [x] Organise the documentation

## To Test

```bash
doxygen Doxyfile 
cd docs
make html && python -m http.server -d build/html
```

Fixes # (issue)

## Before submitting


sovrasov:
@ashwinvaidya17 thanks for the prompt PR! Could you fetch the latest master into your branch? Does the code style failure look reasonable?

ashwinvaidya17:
It flags the make.bat file. I didn't write it. It was autogenerated by sphinx. My next goal is to slowly add pre-commit checks across the repo.

sovrasov:
@eugene123tw could you have a look at the docs update?

docs/source/python/models/ssd.md
```diff
@@ -0,0 +1,8 @@
+# Ssd
```
eugene123tw, line 1: @sovrasov just for reference. I think we could move `ssd.py` to `detection_model.py` in the future.

docs/source/guides/index.md
```diff
@@ -0,0 +1,9 @@
+# Guides
+
+```{toctree}
+:caption: Guides
+:hidden:
+
+./model-configuration
+./releasing-to-PyPI
```
eugene123tw, line 8: @ashwinvaidya17 @sovrasov I'm not sure if we need to include `PyPI releasing guide` to public doc. What do you think?

eugene123tw (approved):
@ashwinvaidya17 Thanks, Ashwin! I’ve added a few minor comments, but overall, it looks great! We can start building on and enhancing our documentation now :+1:

docs/source/guides/index.md
ashwinvaidya17: I kept the existing docs as is. Ideally we shouldn't have this guide in the first place. Releasing to PyPI can be easily be automated with github actions.

sovrasov: Right, that's only for internal usage

docs/source/python/models/ssd.md
sovrasov: I see your point, but not 100% sure it makes sense to have both abstract `DetectionModel` and inherited from it `SSD` in one file. In opposite, in the future we might want to add other detection models. At the first iteration let's try to save compatibility of imports/includes

PR #209, opened 2024-10-24
Cleanup outdated OMZ wrappers

sovrasov:
# What does this PR do?

- YOLO is not touched, need to clarify the status


eunwoosh (approved):

eugene123tw (approved):
Thanks, Vlad, for cleaning up the PR! I think in the future, we could consider changing the name `MaskRCNNModel` to `InstanceSegmentationModel`, as `MaskRCNNModel` doesn't only support MaskRCNN; it's a general instance segmentation interface that runs post-processing on bboxes, labels, and masks output from OV.

PR #211, opened 2024-10-28
Disable MacOS OVMS testing

sovrasov:
# What does this PR do?

OVMS client fails a gRPC call by timeout in MacOS env. Previously, the test was working, but now it's broken probably because of OVMS upgrade. An attempt to restore it will be carried out after merging https://github.com/openvinotoolkit/model_api/pull/206
Ubuntu 22.04 will be used instead


.github/workflows/test_precommit.yml
```diff
@@ -148,13 +148,11 @@ jobs:
...
-        os: [ubuntu-20.04, ubuntu-22.04, macos-13]
+        os: [ubuntu-20.04, ubuntu-22.04, ubuntu-24.04]
         python-version: [3.9, '3.10', '3.11']
         exclude:
-          - os: macos-13
-            python-version: '3.10'
-          - os: macos-13
-            python-version: '3.11'
+          - os: ubuntu-24.04
+            python-version: '3.9'
```
eunwoosh, line 155: It's just a question. Why is it excluded?

eunwoosh (approved):

.github/workflows/test_precommit.yml
sovrasov: It's unrealistic somebody would use 3.9 with the latest Ubuntu version

PR #222, opened 2024-10-31
Update results classes

sovrasov:
# What does this PR do?

- Don't use untyped `namedtuple`
- Move result entities to a standalone file


model_api/python/model_api/models/result_types.py
```diff
@@ -0,0 +1,247 @@
+#
```
ashwinvaidya17, line 1: Something for future, but maybe we can split this under `types/anomaly.py, types/classification.py`

ashwinvaidya17 (approved):
Thanks for the refactor.

model_api/python/model_api/models/result_types.py
sovrasov: Yeah, that's a possible option

PR #223, opened 2024-10-31
✨Add model descriptions and documentation for instance segmentation, detection, and keypoint detection

eugene123tw:
# What does this PR do?

Add model descriptions and documentation for instance segmentation, detection, and keypoint detection

## Before submitting


docs/source/python/models/detection_model.md
```diff
@@ -1,5 +1,46 @@
...
+
+## OpenVINO Model Specifications
+
+### Inputs
+
+A single input image of shape (H, W, 3) where H and W are the height and width of the image, respectively.
+
+### Outputs
+
+Detection model outputs a list of detection objects (i.e `list[Detection]`) wrapped in `DetectionResult`, each object containing the following attributes:
```
sovrasov, line 15: Good to have this description here, but it seems like a bit of headers reshape is required here.
The initial header says that's an OV model I/O description, which means here should go requirements to the raw OV model's output (for instance `model_api/python/docs/keypoint_detection.md`).
Wrapper I/O specification is extremely useful, and it can be located here under another header.

sovrasov: The goal is to make both OV and Wrapper I/O description more clear, and enable quick adaptation of 3rd party models to be compatible with MAPI. Another obstacle here is rt_info: now only OTX know the magic to generate it, but that also is to be changed.

PR #229, opened 2024-11-07
Type annotation

sovrasov:
# What does this PR do?

- Enables type checking section rule
- Annotate the majority of modules


model_api/python/model_api/adapters/onnx_adapter.py
```diff
@@ -122,6 +124,9 @@ def await_all(self):
     def await_any(self):
         pass
 
+    def get_raw_result(self, infer_result):
```
ashwinvaidya17, line 127: We need to annotate these as well but it something for later when we refactor the result types.

ashwinvaidya17 (approved):

PR #257, opened 2025-01-21
Add Anomaly scene

ashwinvaidya17:
# What does this PR do?

- Add Anomaly Scene

```python
image = Image.open("./data/coco128/images/train2017/000000000074.jpg")
image = image.resize((255, 255))  # Ideally this should be within visualization

heatmap = np.zeros(image.size, dtype=np.uint8)
heatmap[64:96, 64:96] = 255

mask = np.zeros(image.size, dtype=np.uint8)
mask[32:96, 32:96] = 255
mask[40:80, 0:128] = 255

anomaly_result = AnomalyResult(
    anomaly_map=heatmap,
    pred_boxes=np.array([[0, 0, 128, 128], [32, 32, 96, 96]]),
    pred_label="Anomaly",
    pred_mask=mask,
    pred_score=0.85,
)
visualizer = Visualizer()
visualizer.save(image, anomaly_result, "anomaly_result.jpg")
```

image

```python
image = Image.open("./data/coco128/images/train2017/000000000074.jpg")
image_array = cv2.cvtColor(np.array(image), cv2.COLOR_RGB2BGR)

padim = AnomalyDetection.create_model(
    "./data/otx_models/anomaly_stfpm_bottle_mvtec.xml"
)

predictions: AnomalyResult = padim(image_array)
visualizer = Visualizer()
visualizer.save(image, predictions, "visualizer_output.jpg")
```

image

Fixes # (issue)

## Before submitting

### DIFF
--- a/src/python/model_api/visualizer/primitive/label.py
+++ b/src/python/model_api/visualizer/primitive/label.py
@@ -40,13 +40,13 @@ class Label(Primitive):
 
     def __init__(
         self,
-        label: str,
+        label: Union[str, float],
         fg_color: Union[str, tuple[int, int, int]] = "black",
         bg_color: Union[str, tuple[int, int, int]] = "yellow",
         font_path: Union[str, BytesIO, None] = None,
         size: int = 16,
     ) -> None:
-        self.label = label
+        self.label = str(label)
         self.fg_color = fg_color
         self.bg_color = bg_color
         self.font = ImageFont.load_default(size=size) if font_path is None else ImageFont.truetype(font_path, size)
--- a/src/python/model_api/visualizer/primitive/label.py
+++ b/src/python/model_api/visualizer/primitive/label.py
@@ -19,6 +19,7 @@ class Label(Primitive):
 
     Args:
         label (str): Text of the label.
+        score (float | None): Score of the label. This is optional.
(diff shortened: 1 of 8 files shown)


sovrasov:
@ashwinvaidya17 given the real-word example it looks like we need to introduce confidence rounding parameter to avoid long labels representations.

src/python/model_api/visualizer/primitive/label.py
```diff
@@ -40,13 +40,13 @@ class Label(Primitive):
...
     def __init__(
         self,
-        label: str,
+        label: Union[str, float],
         fg_color: Union[str, tuple[int, int, int]] = "black",
         bg_color: Union[str, tuple[int, int, int]] = "yellow",
         font_path: Union[str, BytesIO, None] = None,
         size: int = 16,
     ) -> None:
-        self.label = label
```
sovrasov, line 49: Why do we need to construct label from `float`?

ashwinvaidya17: This was in case someone passes scores instead of text. One option is that the users convert the scores to string before passing it to the class. Or, we can let the class handle this. If we let the class handle this then we can introduce confidence rounding. Let me know what you prefer.

sovrasov (approved):

PR #258, opened 2025-01-22
CVS-160560 nanobindings for keypoint detection and segmentation

mgumowsk:
# What does this PR do?

Fixes # (issue)

## Before submitting


src/cpp/py_bindings/py_instance_segmentation.cpp
```diff
@@ -0,0 +1,76 @@
...
+                 std::vector<ImageInputData> input_mats;
+                 input_mats.reserve(inputs.size());
+
+                 for (const auto& input : inputs) {
+                     input_mats.push_back(pyutils::wrap_np_mat(input));
+                 }
+
+                 return self.inferBatch(input_mats);
+             })
+        .def("postprocess",
```
sovrasov, line 54: Do we really need to expose this method now? `InferenceResult` is hard to arrange from python, maybe I don't fully understand the usecase though

mgumowsk: removed

src/cpp/py_bindings/py_instance_segmentation.cpp
```diff
@@ -0,0 +1,72 @@
...
+                 return self.inferBatch(input_mats);
+             })
+        .def_prop_ro_static("__model__", [](nb::object) {
+            return MaskRCNNModel::ModelType;
+        });
+
+    nb::class_<InstanceSegmentationResult, ResultBase>(m, "InstanceSegmentationResult")
+        .def(nb::init<int64_t, std::shared_ptr<MetaData>>(), nb::arg("frameId") = -1, nb::arg("metaData") = nullptr)
+        .def_prop_ro(
+            "feature_vector",
```
sovrasov, line 61: Iseg result also contains a saliency map as a vector of `cv::Mat`. Perhaps, it's not that easy to expose but there is a chance a vector of `nb::ndarray<uint8_t...` would work here

src/cpp/py_bindings/py_anomaly.cpp
```diff
@@ -0,0 +1,80 @@
...
+        .def_prop_ro(
+            "anomaly_map",
+            [](AnomalyResult& r) {
+                return nb::ndarray<uint8_t, nb::numpy, nb::c_contig>(r.anomaly_map.data,
+                                                                     {static_cast<size_t>(r.anomaly_map.rows),
+                                                                      static_cast<size_t>(r.anomaly_map.cols),
+                                                                      static_cast<size_t>(r.anomaly_map.channels())});
+            },
+            nb::rv_policy::reference_internal)
+        .def_ro("pred_boxes", &AnomalyResult::pred_boxes)
```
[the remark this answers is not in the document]
mgumowsk: Added wrapping and tested it works.

src/cpp/py_bindings/py_segmentation.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+                return nb::ndarray<float, nb::numpy, nb::c_contig>(ir.feature_vector.data(),
+                                                                   ir.feature_vector.get_shape().size(),
+                                                                   ir.feature_vector.get_shape().data());
+            },
+            nb::rv_policy::reference_internal)
+        .def_prop_ro(
+            "soft_prediction",
+            [](ResultBase& r) {
+                ImageResultWithSoftPrediction ir = r.asRef<ImageResultWithSoftPrediction>();
```
sovrasov, line 85: Would that work if the result entity is actually an `ImageResult`?
In sseg wrapper now we have the following logic:
```
if (return_soft_prediction) {
  return a ptr to ImageResultWithSoftPrediction
}
else 
  return a ptr to ImageResult
```
`infer()` method's return type is a ptr to `ImageResult`. I think that kind of design looks weird an overall inconvenient, because we always need to check if type conversion to `ImageResultWithSoftPrediction` is successful.
Related refactoring can be done in subsequent PR, here we just need to be sure that both the cases `return_soft_prediction==true/false` are handled correctly

PR #278, opened 2025-04-02
Merge ModelBase into ImageModel

RHeckerIntel:
First step in attempting to reduce complexity. 

First by adding complexity by merging the model base and image model. The second step step would be to extract as much of the logic and properties into separate modules.


src/cpp/models/include/models/image_model.h
```diff
@@ -20,7 +23,7 @@ struct InputData;
 struct InternalModelData;
 
 // ImageModel implements preprocess(), ImageModel's direct or indirect children are expected to implement prostprocess()
-class ImageModel : public ModelBase {
+class ImageModel {
```
sovrasov, line 26: Can we rename ImageModel -> Model, since that's the only base model now?

sovrasov (approved):

PR #304, opened 2025-06-13
Refactor - Implement Classification

### LINKED ISSUE #292: Refactor semantic segmentation wrapper
Implement serialization for semantic segmentation

### LINKED ISSUE #293: Implement instance segmentation serialization
Instance segmentation itself is in there. Just serialization step and removing the gtest skip from test_accuracy test.

### LINKED ISSUE #294: Implement Classification Model
Implement Classification model for new structure:

1. Implement basic model
2. implement serialization step

RHeckerIntel:
Implements Classification Model into the refactor feature branch

Also managed to fix issue that the serialization gives slightly different output. The cause was a missing ov::set_batch(model, 1);
This mirrored the issue in instance and semantic segmentation and so also enabled those tests

Closes #294 
Closes #293 
Closes #292


src/cpp/include/tasks/classification.h
```diff
@@ -0,0 +1,95 @@
...
+#include "tasks/results.h"
+#include "utils/config.h"
+#include "utils/vision_pipeline.h"
+
+class Classification {
+public:
+    std::shared_ptr<InferenceAdapter> adapter;
+    VisionPipeline<ClassificationResult> pipeline;
+
+    Classification(std::shared_ptr<InferenceAdapter> adapter, cv::Size input_shape)
```
sovrasov, line 22: Does it make sense to define a constant like `DEFAULT_INPUT_SHAPE`? Or maybe hide this parameter: I'd opt for avoiding exposing model reshaping feature for now. Only a fraction of Geti models support  reshaping, let alone arbitrary public models

RHeckerIntel: Yes, good point. The problem is that its model dependent. But I believe its added during serialization step, so I can just retrieve it again. I was just solving a variable scope issue, but I'll figure it out and apply everywhere.

sovrasov (approved):

PR #310, opened 2025-06-18
Support ONNX models via OV

### LINKED ISSUE #308: Support ONNX models via OV
MAPI should consume ONNX as well. OV can process these models, but a refactoring to eliminate explicit om model usage is required to properly abstract model metadata

sovrasov:
# What does this PR do?
Port of python implementation:
- onnx model is not serializable, pre-processing can be embedded on-the -fly only
- onnx metadata is used to store rt_info configuration
- some of the onnx tests are restored
- onnxrt cmake integration sucks

resolves #308

Fixes # (issue)

## Before submitting

- [x] Did you write any new necessary tests?


src/cpp/CMakeLists.txt
```diff
@@ -30,3 +36,11 @@ add_library(model_api STATIC ${TASK_SOURCES} ${TASKS_SOURCES} ${UTILS_SOURCES} $
...
 target_link_libraries(model_api PUBLIC openvino::runtime opencv_core opencv_imgproc PRIVATE nlohmann_json::nlohmann_json)
 target_include_directories(model_api PUBLIC ${PROJECT_SOURCE_DIR}/include)
+
+# onnxrt doesn't have a working CMake config, set the related directories manually
+set(ONNXRUNTIME_ROOTDIR ${CMAKE_BINARY_DIR}/_deps/onnxrt-src)
+
+target_include_directories(model_api PUBLIC ${ONNXRUNTIME_ROOTDIR}/include)
+
+target_link_directories(model_api PUBLIC "${ONNXRUNTIME_ROOTDIR}/lib")
+target_link_libraries(model_api PUBLIC libonnxruntime.so)
```
RHeckerIntel, line 46: This is rather platform specific, should we try to install onnxruntime using install_dependencies instead?

I've installed the same binary downloaded above and it has working cmake for me.

src/cpp/src/tasks/detection/ssd.cpp
```diff
@@ -115,8 +115,10 @@ void SSD::serialize(std::shared_ptr<ov::Model> ov_model) {
         // prepareMultipleOutputs(ov_model); //This does nothing from what I can see.
     }
 
-    ov_model->set_rt_info(input_shape[0], "model_info", "orig_width");
-    ov_model->set_rt_info(input_shape[1], "model_info", "orig_height");
+    config["orig_width"] = std::to_string(input_shape[0]);
+    config["orig_height"] = std::to_string(input_shape[1]);
+
+    return config;
```
RHeckerIntel, line 121: we need to add the rt info for serialization step anyway. I'm not sure I like adding the config and returning it. I dont like retrieving the config twice too, but it does feel neat to have serialize as much as possible a separate step. Aside fro the orig_width/height what is the reason for returning the config?

src/cpp/CMakeLists.txt
sovrasov: Good to hear cmake is working in your case. Could you give me a hint how did you manage to use it? `add_directory`?

sovrasov: Updated cmake to support Linux for now, platforms list is extendable

RHeckerIntel (approved):
awesome!

PR #538, opened 2026-04-17
feat: intensity scaling

### LINKED ISSUE #520: # Add intensity scaling and input dtype support to ModelAPI
# Add intensity scaling and input dtype support to ModelAPI

## Problem

ModelAPI hardcodes `uint8` assumptions in preprocessing:

- `embed_preprocessing` defaults `dtype=int` → `Type.u8`; no way to signal f32 or u16 from IR metadata.
- Pad constants in `resize_image_with_aspect_graph` / `resize_image_letterbox_graph` are `np.uint8` — causes OV validation errors when the input tensor is f32 or u16.
- `pad_value` is capped at [0, 255].

### 1. New parameters in `IMAGE_PREPROCESSING`

```python
"input_dtype": StringValue(default_value="u8", choices=("u8", "f32", "u16")),
"intensity_mode": StringValue(default_value="none", choices=("none", "scale_to_unit", "window", "range_scale")),
"intensity_max_value": NumericalValue(float, default_value=None, min=0.0),
"intensity_window_center": NumericalValue(float, default_value=None),
"intensity_window_width": NumericalValue(float, default_value=None),
"intensity_scale_factor": NumericalValue(float, default_value=1.0),
"intensity_min_value": NumericalValue(float, default_value=0.0),
```

Auto-read from `rt_info["model_info"]` via `Model._load_config`.

### 2. Plumb `input_dtype` through `ImageModel` → `embed_preprocessing`

Map `input_dtype` to the `dtype` kwarg (`"u8"` → `int`, `"f32"` → `float`, `"u16"` → `Type.u16`).

### 3. Intensity preprocessing nodes in `embed_preprocessing`

Before resize and mean/scale, insert OV nodes per `intensity_mode`:

| Mode | Graph operation |
|------|-----------------|
| `scale_to_unit` | `divide(input, max_value)` → `convert(f32)` |
| `window` | `subtract(low)` → `divide(high - low)` → `clamp(0, 1)` |
| `range_scale` | `multiply(factor)` → `clamp` → `subtract(min)` → `divide(max - min)` |

### 4. Fix hardcoded `np.uint8` pad constants

Infer element type from the input node instead of hardcoding `np.uint8`. Widen `pad_value` max to 65535.

### 5. Python-side fallback

Thread `intensity_mode` through `InputTransform` and `setup_python_preprocessing_pipeline` for NPU/ONNX paths.

## What this unblocks

- OTX and Geti delete their adapter subclasses and workarounds
- uint16 inference (thermal, medical) works end-to-end
- Standalone ModelAPI users get correct preprocessing for any exported OTX model

mgumowsk:
# What does this PR do?

Fixes #520

## Before submitting

### DIFF
--- a/src/model_api/adapters/inference_adapter.py
+++ b/src/model_api/adapters/inference_adapter.py
@@ -216,6 +216,15 @@ class InferenceAdapter(ABC):
         mean: list[Any] | None = None,
         scale: list[Any] | None = None,
         input_idx: int = 0,
+        input_dtype: str = "u8",
+        intensity_mode: str = "none",
+        intensity_max_value: float | None = None,
+        intensity_window_center: float | None = None,
+        intensity_window_width: float | None = None,
+        intensity_percentile_low: float = 1.0,
+        intensity_percentile_high: float = 99.0,
+        intensity_scale_factor: float = 1.0,
+        intensity_min_value: float = 0.0,
     ):
         """
         Embeds preprocessing into the model if possible with the adapter being used.
@@ -234,4 +243,13 @@ class InferenceAdapter(ABC):
             mean (list[Any] | None, optional): Mean values to perform input normalization. Defaults to None.
             scale (list[Any] | None, optional): Scale values to perform input normalization. Defaults to None.
             input_idx (int, optional): Index of the model input to apply preprocessing to. Defaults to 0.
+            input_dtype (str, optional): Input element type string ("u8", "f32", "u16"). Defaults to "u8".
+            intensity_mode (str, optional): Intensity scaling mode. Defaults to "none".
+            intensity_max_value (float | None, optional): Max value for intensity scaling. Defaults to None.
+            intensity_window_center (float | None, optional): Window center for window mode. Defaults to None.
+            intensity_window_width (float | None, optional): Window width for window mode. Defaults to None.
+            intensity_percentile_low (float, optional): Low percentile for percentile mode. Defaults to 1.0.
+            intensity_percentile_high (float, optional): High percentile for percentile mode. Defaults to 99.0.
+            intensity_scale_factor (float, optional): Scale factor for range_scale mode. Defaults to 1.0.
+            intensity_min_value (float, optional): Min value for range_scale mode. Defaults to 0.0.
         """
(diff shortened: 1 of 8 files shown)


tybulewicz (approved):

kprokofi (commented):
1) OTX writes ("model_info", "intensity_repeat_channels") when [repeat_channels > 0], but model_api has no corresponding parameter in ParameterRegistry. This means single-channel→3ch expansion for pretrained backbones won't happen at inference. OTX's RepeatChannels transform does this during training, but the exported IR won't reproduce it via model_api. The purpose is: single-channel images (grayscale medical, thermal) need to be expanded to 3 channels because pretrained backbones (ResNet, EfficientNet, etc.) expect 3-channel RGB input. The default should be False or 0 (meaning "don't repeat"). The logic is:
 * 0 / False → no-op, pass through as-is (for normal 3-channel RGB images)
 * 3 / True → repeat single-channel to 3 channels (for grayscale/medical/thermal)
The exporter only writes the key when repeat_channels > 0, so for standard uint8 RGB models this metadata is absent, and model_api should default to False (no repetition) — preserving backward compatibility.
2) OTX's RangeScale does: (clamp(x * scale_factor, min, max) - min) / (max - min) → output in [0, 1].
model_api's _range_scale in create_intensity_fn does: clip(x * scale_factor, min, max) — no subtraction and division by range. The OV graph version (range_scale_preprocess_graph) similarly just clamps without normalizing. This means range_scale output won't be [0, 1] and subsequent mean/scale normalization will produce wrong results. It is critical issue.
3) Missing i16 (int16) support in model_api. OTX's _dtype_map includes "int16": "i16", but model_api's input_dtype StringValue only allows ("u8", "f32", "u16"). If OTX exports a model with int16 storage, model_api will reject the i16 value. The _DTYPE_TO_OV map in openvino_adapter.py also lacks i16 → Type.i16. 
int16 appears in some medical imaging formats (DICOM CT scans store Hounsfield units as signed 16-bit integers, e.g., -1024 to +3071).

mgumowsk:
Added task https://github.com/open-edge-platform/model_api/issues/540

PR #548, opened 2026-04-27
feat: intensity scaling extend functionality

### LINKED ISSUE #540: intensity scaling: extend functionality with additional parameters
During review of https://github.com/open-edge-platform/model_api/pull/538 additional scope requested:

1. OTX writes ("model_info", "intensity_repeat_channels") when [repeat_channels > 0], but model_api has no corresponding parameter in ParameterRegistry. This means single-channel→3ch expansion for pretrained backbones won't happen at inference. OTX's RepeatChannels transform does this during training, but the exported IR won't reproduce it via model_api. The purpose is: single-channel images (grayscale medical, thermal) need to be expanded to 3 channels because pretrained backbones (ResNet, EfficientNet, etc.) expect 3-channel RGB input. The default should be False or 0 (meaning "don't repeat"). The logic is:

- 0 / False → no-op, pass through as-is (for normal 3-channel RGB images)

- 3 / True → repeat single-channel to 3 channels (for grayscale/medical/thermal)

The exporter only writes the key when repeat_channels > 0, so for standard uint8 RGB models this metadata is absent, and model_api should default to False (no repetition) — preserving backward compatibility.

2. OTX's RangeScale does: (clamp(x * scale_factor, min, max) - min) / (max - min) → output in [0, 1].
model_api's _range_scale in create_intensity_fn does: clip(x * scale_factor, min, max) — no subtraction and division by range. The OV graph version (range_scale_preprocess_graph) similarly just clamps without normalizing. This means range_scale output won't be [0, 1] and subsequent mean/scale normalization will produce wrong results. It is critical issue.

3. Missing i16 (int16) support in model_api. OTX's _dtype_map includes "int16": "i16", but model_api's input_dtype StringValue only allows ("u8", "f32", "u16"). If OTX exports a model with int16 storage, model_api will reject the i16 value. The _DTYPE_TO_OV map in openvino_adapter.py also lacks i16 → Type.i16.
int16 appears in some medical imaging formats (DICOM CT scans store Hounsfield units as signed 16-bit integers, e.g., -1024 to +3071).

mgumowsk:
# What does this PR do?

Fixes #540

- add intensity_repeat_channels
- clamps range_scale in [0,1]
- add support for int16

### DIFF
--- a/src/model_api/adapters/openvino_adapter.py
+++ b/src/model_api/adapters/openvino_adapter.py
@@ -39,6 +39,7 @@ from .utils import (
     get_rt_info_from_dict,
     load_parameters_from_onnx,
     range_scale_preprocess,
+    repeat_channels_preprocess,
     resize_image,
     resize_image_letterbox,
     resize_image_with_aspect,
@@ -417,6 +418,7 @@ class OpenvinoAdapter(InferenceAdapter):
         intensity_percentile_high: float = 99.0,
         intensity_scale_factor: float = 1.0,
         intensity_min_value: float = 0.0,
+        intensity_repeat_channels: bool = False,
         input_frame_shape: tuple[int, int] | None = None,
     ) -> None:
         """
@@ -444,6 +446,7 @@ class OpenvinoAdapter(InferenceAdapter):
                 intensity_percentile_high=intensity_percentile_high,
                 intensity_scale_factor=intensity_scale_factor,
                 intensity_min_value=intensity_min_value,
+                intensity_repeat_channels=intensity_repeat_channels,
             )
             self.use_python_preprocessing = True
             input_name = self.model.inputs[input_idx].get_any_name()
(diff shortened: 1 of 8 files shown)


tybulewicz (dismissed):

src/model_api/adapters/openvino_adapter.py
```diff
@@ -456,17 +459,23 @@ def embed_preprocessing(
...
 
-        ppp.input(input_idx).tensor().set_layout(ov.Layout("NHWC")).set_color_format(
-            ColorFormat.BGR,
-        )
+        ppp.input(input_idx).tensor().set_layout(ov.Layout("NHWC"))
+        if not intensity_repeat_channels:
+            ppp.input(input_idx).tensor().set_color_format(ColorFormat.BGR)
+
+        # Handle repeat single-channel to 3 channels
+        if intensity_repeat_channels:
```
kprokofi, line 472: else?
```suggestion
       else:
```

src/model_api/adapters/utils.py
```diff
@@ -753,6 +780,18 @@ def change_layout(image, layout):
 }
 
 
+def _repeat_single_channel_np(image: np.ndarray) -> np.ndarray:
```
kprokofi, line 783: Duplicate from  image_model.py?

src/model_api/adapters/openvino_adapter.py
mgumowsk: fixed

src/model_api/adapters/utils.py
mgumowsk: Yes, good catch. I moved this function to the utils, but forgot to remove it from image_model. Fixed.

kprokofi (approved):

PR #551, opened 2026-04-29
feat: restructure repository in model_api and model_converter projects

### LINKED ISSUE #508: Restructure repo into two independent subprojects: `model_api` and `model_converter`
We want to give `model_api` and `model_converter` their own independent dependency sets and packaging making it easier to consume just the inference library without pulling in conversion-time dependencies.

```
model_api/
├── src/model_api/      ← importable package
├── tests/
├── docs/
├── pyproject.toml      ← only: numpy, opencv, openvino, pillow
└── uv.lock
model_converter/
├── src/model_converter/  ← importable package
├── tests/
├── pyproject.toml        ← torch, torchvision, openvino, onnx, etc.
└── uv.lock
```

Shared repo-level files (LICENSE, README, CI workflows) stay at the root.

Each subproject gets its own pyproject.toml with scoped dependencies and a uv.lock. As part of this, [project.optional-dependencies] (tests, docs) would be migrated to [dependency-groups], which uv supports natively.

mgumowsk:
# What does this PR do?

Fixes #508


tybulewicz:
Please rebase this repo and run `ruff` again to re-format files.

tybulewicz (dismissed):

.github/workflows/pre_commit.yml
```diff
@@ -35,7 +35,9 @@ jobs:
           version: "0.11.3"
 
       - name: Install dependencies
-        run: uv sync --locked --all-extras
+        run: |
+          uv --directory model_api sync --locked --all-extras --group docs
+          uv --directory model_converter sync --locked
```
ivanzati, line 40: Perhaps sync for model_converter is not mandatory as workflow goes through model_api files only

mgumowsk: True, not required for now, but it's just the first step. The next step is adding checks and tests for model_converter.

ivanzati (dismissed):
Thank you, Mariusz!

ivanzati (dismissed):
Thank you, Mariusz!

model_api/pyproject.toml
```diff
@@ -37,36 +37,35 @@ dependencies = [
 onnx = [
```
mpryahin, line 37: could it be possible to pin dependencies?

model_converter/pyproject.toml
```diff
@@ -0,0 +1,198 @@
...
+description = "Model Converter: tools for converting PyTorch models to OpenVINO IR"
+readme = "README.md"
+classifiers = [
+    "License :: OSI Approved :: Apache Software License",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+]
+dependencies = [
```
mpryahin, line 29: could it be possible to pin dependencies?

CONTRIBUTING.md
```diff
@@ -35,17 +35,24 @@ Set up your development environment to start contributing. This involves install
 <details>
 <summary>Development Environment Setup Instructions</summary>
 
-1. Create a new python environment and install the development requirements:
+1. Create the subproject environments and install the development requirements:
```
mpryahin: Do we always need to install all dependencies for both projects, or are there cases where a developer only needs one of them? For example, when running tests for just a single project.

.github/workflows/docs.yml
```diff
@@ -31,11 +31,11 @@ jobs:
 
       - name: Install dependencies
         run: |
-          uv sync --locked --extra docs
+          uv --directory model_api sync --locked --group docs
```
mpryahin, line 34: Maybe we could parameterise this workflow instead of explicitly tying it to model-api, and reuse it across both projects

mpryahin (commented):
Thanks Mariusz, just a couple of minor comments

model_converter/pyproject.toml
mgumowsk: Unless there's a specific reason for a given library, like a bug or security issue, then we can pin dependency. The specific versions are pinned in uv.lock

.github/workflows/docs.yml
mgumowsk: Definitely, I already did that in other workflows. This one is only for model_api, because we don't have any docs for model_converter yet. Once we have docs in model_converter, this will be parametrisied.

CONTRIBUTING.md
mgumowsk: No need to install both. I'll rewrite this document, and add separete sections for model_api and model_converter.

tybulewicz (dismissed):

mpryahin (approved):
Thanks Mariusz!

AlexanderBarabanov (approved):

PR #587, opened 2026-05-19
Disable uv upgrade by renovate bot

ivanzati:
# What does this PR do?

Disable uv upgrade by renovate bot and fix version of renovate in a way that it can update itself.

## Before submitting

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -32,6 +32,23 @@
       schedule: ["* * 1 * *"],
       matchPackageNames: ["*"],
     },
+    // uv version used in GitHub Actions is updated manually
+    {
+      enabled: false,
+      matchDatasources: ["github-releases"],
+      matchDepNames: ["astral-sh/uv"],
+      matchDepTypes: ["uses-with"],
+    },
+  ],
+  // Is used to upgrade Renovate version
+  customManagers: [
+    {
+      fileMatch: ["^\\.github/workflows/[^/]+\\.ya?ml$"],
+      matchStrings: [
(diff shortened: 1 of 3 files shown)


.github/renovate.json5
```diff
@@ -32,6 +32,23 @@
...
+      matchDepTypes: ["uses-with"],
+    },
+  ],
+  // Is used to upgrade Renovate version
+  customManagers: [
+    {
+      fileMatch: ["^\\.github/workflows/[^/]+\\.ya?ml$"],
+      matchStrings: [
+        "# renovate: datasource=(?<datasource>.*?) depName=(?<depName>.*?)\\s+.+version: (?<currentValue>.*)",
+        "# renovate: datasource=(?<datasource>.*?)\\s+export RENOVATE_IMAGE=(?<depName>[^:]+):(?<currentValue>[^@]+)(?:@(?<currentDigest>sha256:[a-f0-9]+))?",
```
AlexanderBarabanov, line 49: Please add sha into https://github.com/open-edge-platform/model_api/blob/master/.github/workflows/renovate-config-validator.yml#L45 as this regexp may not work for the current state.

.github/renovate.json5
AlexanderBarabanov: You need to explicitly enable `custom.regex` manager here to support added regexp.
See example in https://github.com/openvinotoolkit/physicalai/blob/main/.github/renovate.json5#L40

AlexanderBarabanov:
```suggestion
      matchManagers: ["github-actions", "custom.regex"],
```

Otherwise this new regexp will not work

.github/renovate.json5
ivanzati: Thanks, will add!

AlexanderBarabanov (approved):

PR #591, opened 2026-05-20
Enable traffic rules with harden-runner

ivanzati:
# What does this PR do?

Enable traffic rules with harden-runner step in workflows. The PyPi publish and sbom collect workflows are skipped at the moment until their pattern will be known later.

## Before submitting


AlexanderBarabanov (approved):
Thank you!

PR #645, opened 2026-08-18
feat: visualization custom label colours

### LINKED ISSUE #638: Extend visualization to support label colours
The label colour used during visualization is arbitrary, and may not match the colour that the user expects (such as from their project https://github.com/open-edge-platform/geti/pull/7040).

A solution for this would be to extend the visualization interface with an optional colour mapping.

mgumowsk:
# What does this PR do?

Fixes #638

## Before submitting


tybulewicz (changes requested):
For MaskRCNN models (for example model_data19) visualization with tupples fails with `AttributeError: 'tuple' object has no attribute 'lower'`.
Please make sure to add a unit test for this.

tybulewicz (approved):
LGTM

PR #652, opened 2026-09-11
Support dynamic batch dimension in model converter

warrkan:


model_converter/src/model_converter/converters/pytorch.py
```diff
@@ -175,8 +175,11 @@ def export_to_openvino(
             model = self._prepare_model_for_export(model, model_config)
             model.eval()
             dummy_input = self._create_example_input(input_shape, model_config)
+            # Keep spatial dims static, but let the batch dimension be dynamic so any batch size works at inference.
+            dynamic_shape = ov.PartialShape([-1, *input_shape[1:]])
```
kprokofi: I would rather make it generic and expose a dynamic batch parameter 
Or more over "batch_size" where "-1" will be dynamic and specific value - desired fixed batch size for the export

kprokofi (approved):

